PR #7, opened 2019-06-05
Use husky to manage deps

damccorm:
I think we can use husky to easily manage what needs to be checked in. It uses git hooks to require that this stuff is run before checkin. This way we also don't need to manually keep track of packages in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
+!node_modules/
 __tests__/runner/*


ONFU:
Live message at Nerdeir.Peter.Kozlov.666.69@hotmail.com am valkhylmher valentine valkirye vendetta l need the disturb of the NoX-PAcT am general of midle earth philipine NOX-PacT lts woste of the districk gestioning all gouvernement all the World lt's the reddick commence of the districk health

ONFU:
the adjudent chef of Operation of NoX-PAcT comandement Operation will be answer whith message will be follow laurence trhynpanier and mme eves Shafner districk of 
NoX-PAcT

PR #28, opened 2019-08-12
Use node-version instead of version

damccorm:


soerface:
Is it possible to make pull requests for the GitHub help pages? Just saw that here, the deprecated "version" instead of "node-version" is used, I would like to update it: https://help.github.com/en/articles/workflow-syntax-for-github-actions#example-7

damccorm:
I don't believe that repo is open sourced to avoid confusion since it has docs for non-released products. I added a PR there though.

soerface:
Thanks! Can you fix the lonely `$`, too? It needs to be something like `${{ matrix.node-version }}` / `${{ matrix.os }}`. Is there a place to report errors in the GitHub help, so I won't need to continue bothering you in this pull request? ;)

damccorm:
> Can you fix the lonely $, too?

Yep

> Is there a place to report errors in the GitHub help, so I won't need to continue bothering you in this pull request?

I'm actually not 100% sure where the best place for Action docs bugs is, @rachmari can you help there?

In general, filing an issue in this repo or one of the other https://github.com/actions/ repos will probably get the job done, I'll definitely follow up and PR or route as appropriate.

Thanks for the feedback here as well, definitely helps :)

PR #34, opened 2019-08-18
Use ncc compile source file as a single file

zcong1993:
Use [@zeit/ncc](https://github.com/zeit/ncc) compile source file as a single file.
This can make git repo clean.


damccorm:
I don't think we want this - IMO this actually makes it less clear what is going on in the repo. I think the real solution here is to move to the workflow described [here](https://github.com/actions/toolkit/blob/master/docs/javascript-action.md), where node_modules are only included on release branches

zcong1993:
Why this make less clear ? This solution can ignore node_modules even in release repo. Do you think current the current compiled [lib](https://github.com/actions/setup-node/tree/master/lib) still readable or more clear than this ?

damccorm:
I think the current lib is more readable. It maintains the separation of concerns of different files, and maps to the TypeScript well. This is important so that a reviewer can verify that changes to the TS files match changes in the JS files.

zcong1993:
The compiled files in other static languages such like golang mean the binary file. Does a reviewer need review the binary file changes?

PR #39, opened 2019-08-21
Force all scope names to be lowercased

### LINKED ISSUE #38: Workflow fails to publish to GitHub Package Registry if repo owner login casing does not match scope casing
I've been testing out an Action I was building today and having issues when verifying a step to publish the package to the GitHub Package Registry, which instead attempts to publish the package to the original NPM Registry:

https://github.com/JamesMGreene/test-repo/runs/199833215

After trying to track down the issue for longer than I care to mention as I went down a few incorrect rabbit holes, I believe the root cause is pretty simple: my GitHub username has uppercase letters in it but my NPM scoped package name, by necessity for validation, is all lowercased.

This creates a casing mismatch between the `.npmrc` file contents that get generated by the `setup-node@v1` Action and the scope used by `npm publish` in a later Step. Some highlights from the Workflow Run logs:

Screen Shot 2019-08-21 at 5 43 09 PM

Luckily, the fix for this should also be pretty simple: to lowercase the repo owner's login scope that gets written to the `.npmrc` file. 👌

JamesMGreene:
Fixes #38

### DIFF
--- a/__tests__/__snapshots__/authutil.test.ts.snap
+++ b/__tests__/__snapshots__/authutil.test.ts.snap
@@ -7,12 +7,12 @@ registry=https://registry.npmjs.org/"
 
 exports[`installer tests Automatically configures GPR scope 1`] = `
 "npm.pkg.github.com/:_authToken=\${NODE_AUTH_TOKEN}
-@owner:registry=npm.pkg.github.com/"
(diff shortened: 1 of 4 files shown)


damccorm:
Thanks for tracking this down/the fix - looks great!

PR #40, opened 2019-08-22
Update .gitignore

### LINKED ISSUE #30: Note using recommended gitignore
GitHub publishes a recommended gitignore for NodeJS projects at https://github.com/github/gitignore/blob/master/Node.gitignore

That approach is not used here. Which is fine.

Please reference the best practice used for departing from the above gitignore, preferably at the top of the gitignore here.

damccorm:
GitHub publishes a list of recommended `.gitignore` files, including a node one at https://github.com/github/gitignore/blob/master/Node.gitignore.

This PR aligns this repo more closely with that. If we want this we can port it over to the other setup actions and the template.

Fixes #30


stephenmichaelf:
@damccorm LGTM. Let's check out the other ones and see where we can apply to other setup actions? Can setup issues to track so we don't forget.

PR #42, opened 2019-08-22
Expose the ability to select different architectures

implausible:
Node still supports both x86 and x64 architectures for the Windows operating system. As well, Node is still shipping x86 binaries for the maintenance LTS version 8 for Linux until the end of the year. The GitHub action for setting up Node should support the ability to run x86 variants of Node on an x64 machine.


implausible:
> Why wouldn't you just use a container to run in the correct environment?

I maintain a project called NodeGit, which is a native node module. In it, we distribute precompiled binaries utilizing node-pre-gyp. While the GitHub servers for actions are not x64, utilizing x86 node will force the compilation of the native modules in x86 mode. I also maintain a few other native modules, and while they don't have prebuilts, they need to at least run tests in x86 mode on Windows.

I wish I could tell you it's 2019 and we're done with x86 :), but my primary work is on GitKraken, where we still release and maintain an x86 distributable for Windows using several of these native modules. It would surprise you to know how many people are still using the x86 variant :(.

__tests__/installer.test.ts
```diff
@@ -120,4 +120,30 @@ describe('installer tests', () => {
     await installer.getNode('252');
     await installer.getNode('252.0');
   });
+
+  it('Acquires specified x86 version of node if no matching version is installed', async () => {
```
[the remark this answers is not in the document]
implausible: That's a great idea! Thanks

src/installer.ts
```diff
@@ -81,7 +83,10 @@ export async function getNode(versionSpec: string) {
   core.addPath(toolPath);
 }
 
-async function queryLatestMatch(versionSpec: string): Promise<string> {
+async function queryLatestMatch(
```
damccorm: osArch doesn't need to be added as a param here

implausible: Why is that? It looks like it's used in the switch case immediately below it.

implausible:
@damccorm My apologies for the long wait. I've made the changes you have requested!

src/installer.ts
damccorm: That uses `osPlat`

implausible: Isn't this a use of `os.arch()` that I replaced with `osArch` the parameter?
https://github.com/actions/setup-node/pull/42/files#diff-8c84d074e67a54fc5a18692e0b853feaR159

damccorm: That's a different function (aquireNode vs queryLatestMatch)

bryanmacfarlane:
Thanks.  I think we want this but let me check with product to make sure and if so double check the way it's exposed.

README.md
```diff
@@ -100,6 +100,18 @@ steps:
...
 ```
 
+Specifying a different architecture than the system architecture:
+```yaml
+steps:
+- uses: actions/checkout@master
+- uses: actions/setup-node@v1
+  with:
+    node-version: '10.x'
+    node-arch: 'x86'
```
bryanmacfarlane: As this is an advanced option, I don't want the readme and straight forward example to have it.

bryanmacfarlane:
We can't take binaries as part of an external contribution.  Those have to go through a process.  Please hold.

README.md
bryanmacfarlane: Many folks copy and paste to get started and installing x86 is definately a minor case.  It's ok to note further down in the readme that node-arch is an option and list it's possible values.

src/installer.ts
implausible: Oops, my apologies. Line 94 in the switch case of queryLatestMatch:
https://github.com/actions/setup-node/pull/42/files#diff-8c84d074e67a54fc5a18692e0b853feaR94

implausible:
> We can't take binaries as part of an external contribution. Those have to go through a process. Please hold.

No problem. Right now the binary content is legitimately a text file that's been zipped up. Would this process move quicker if I built the fixtures in the beforeAll of the test suite? @bryanmacfarlane

README.md
implausible: I can understand that.

implausible: I've updated the `README.md` accordingly. Let me know if you have something different in mind.

implausible:
@bryanmacfarlane Here is a patch that could generate the fixtures on each run. Let me know if you would like me to add this patch to this PR.

https://github.com/implausible/setup-node/commit/777d3e4790c96fd0499d7367227a21752f5e5bcc

implausible:
@bryanmacfarlane Was wondering if you had time to take a look at my comment https://github.com/actions/setup-node/pull/42#issuecomment-536641385 here yet.

aminya:
Do we need to redo this pull request or we can rebase?

implausible:
Last time I had taken a look at the rebase, a lot had changed. Probably needs to be redone a little. I can see about putting it on my schedule. No promises for this week or next though.

aminya:
I made a WIP branch here: https://github.com/actions/setup-node/pull/194

implausible:
Thank you @aminya

maxim-lobanov:
Thank you @aminya for great contribution!
@implausible do you have a chance to test it from your side before I update `v2-beta` tag?
You can find examples in usage section in readme. You should specify `actions/setup-node@v2.1.3`

aminya:
I tested it in production. It works without issues:

https://github.com/atom-ide-community/fuzzaldrin-plus-fast/runs/1531634018?check_suite_focus=true#step:5:10

PR #48, opened 2019-08-29
Add setting for always-auth

DavidJFelix:
- https://docs.npmjs.com/misc/config#always-auth
- Allow private repos for stuff like artifactory to work

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -2,6 +2,9 @@ name: 'Setup Node.js environment'
 description: 'Setup a Node.js environment and add it to the PATH, additionally providing proxy support'
 author: 'GitHub'
 inputs:
+  always-auth:
+    description: 'Set always-auth in npmrc'
+    default: 'false'
   node-version:
     description: 'Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0'
     default: '10.x'
--- a/src/authutil.ts
+++ b/src/authutil.ts
@@ -4,14 +4,7 @@ import * as path from 'path';
 import * as core from '@actions/core';
 import * as github from '@actions/github';
 
-interface ConfigureAuthenticationParams {
-  registryUrl: string;
-  alwaysAuth: string;
-}
-export function configAuthentication({
-  registryUrl,
-  alwaysAuth
-}: ConfigureAuthenticationParams) {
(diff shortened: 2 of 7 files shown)


DavidJFelix:
A little background:

JFrog Artifactory (Cloud) wants always-auth=true for both yarn and npm for private repos.
I ran into this issue where auth'd setups would 401 until I made this change.

src/authutil.ts
```diff
@@ -13,10 +20,19 @@ export function configAuthentication(registryUrl: string) {
     registryUrl += '/';
   }
 
-  writeRegistryToFile(registryUrl, npmrc);
+  writeRegistryToFile({registryUrl, fileLocation: npmrc, alwaysAuth});
 }
 
-function writeRegistryToFile(registryUrl: string, fileLocation: string) {
+interface WriteRegistryToFileParams {
```
damccorm: I think I'd rather not define interfaces for these - it feels more natural/readable to me with simple parameters.

action.yml
```diff
@@ -2,6 +2,9 @@ name: 'Setup Node.js environment'
 description: 'Setup a Node.js environment and add it to the PATH, additionally providing proxy support'
 author: 'GitHub'
 inputs:
+  always-auth:
```
damccorm: Nit - this should go after scope in the parameter ordering. That way its grouped with the other auth parameters and comes after registry-url which is required in all auth cases.

src/authutil.ts
DavidJFelix: Yeah I can simplify it. The move to interfaces is just an opinion that parameters should be position independent so that adding new parameters doesn't necessarily affect existing call sites if you add them optionally. Since none of the values felt necessary to the identity of the function, they all felt like they could be ephemeral in the future. Again, just opinion.

action.yml
DavidJFelix: Counter nit: Opinion: At the top it leaves the map in alphabetical order.

DavidJFelix:
I'll make changes as requested. Give me a bit to execute them. I left some justifications in case you're feeling like being persuaded by an internet stranger.

DavidJFelix:
You should see the changes as requested in the most recent commit. Let me know if you want me to rebase them.

PR #50, opened 2019-09-01
Remove unnessary install Yarn command

kidonng:
According to the [docs](https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions), Yarn is installed by default.


damccorm:
Good catch, thanks for the contribution

PR #54, opened 2019-09-04
Deprecate version property

damccorm:
See https://github.com/damccorm/actions-playground/pull/11/checks?check_run_id=212225627 for example output


Terux101:
[bitcoin.pdf](https://github.com/actions/setup-node/files/3885527/bitcoin.pdf)

PR #56, opened 2019-09-10
Example showing how to use this to install private packages securely.

jwalton:
This adds an example to README.md which shows how to use this to install private packages, while also making sure malicious modules with `post-install` scripts can't steal our `NPM_TOKEN`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,24 @@ steps:
     NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
 
+Use private packages:
+```yaml
+steps:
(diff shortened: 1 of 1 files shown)


damccorm:
This looks like a good idea to me, thanks for the contribution!

jwalton:
I suppose splitting up `npm install --no-scripts` and `npm rebuild` into separate things might be overly paranoid.  I was worried that since we had a well defined variable, it would be an easy target for people to steal... But, a malicious package could just read ~/.npmrc right off my drive on my macbook when I `npm install`.  :P

PR #57, opened 2019-09-10
Fix example to use NPM_TOKEN instead of YARN_TOKEN.

jwalton:
Sorry, @damccorm, I found this right after you merged #56.  :P


damccorm:
Oh, good catch. Good to be consistent here

jwalton:
Yeah, I copy-pasted the wrong example to start from.  :P

PR #58, opened 2019-09-13
Add support for node version codenames and keywords

### LINKED ISSUE #26: Support aliases like LTS
Right now, we're thinking it makes sense to support `lts`, `latest`, `current`, and LTS version names (e.g. `dubinium`)

JimiC:
Resolves #26

This PR adds support for node codenames (`argon`, `boron` etc) and selected keywords (`latest`, `current`, `lts`).

Resolving the codenames and keywords is fully automated and ensures future codenames support.

In order to do so, I took the liberty to refactor the code and test cases for better optimization and maintainability.

I'm aware that code reviewing will be a nightmare because of the many changes. I suggest having a look at the new code itself instead, to have a better understanding of it.

In case the refactoring is not acceptable, I'll try to provide a PR with minimum changes, but the code will be ugly.

### DIFF
--- a/__tests__/__snapshots__/authutil.test.ts.snap
+++ b/__tests__/__snapshots__/authutil.test.ts.snap
@@ -1,30 +1,30 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`installer tests Appends trailing slash to registry 1`] = `
+exports[`auth tests Appends trailing slash to registry 1`] = `
 "//registry.npmjs.org/:_authToken=\${NODE_AUTH_TOKEN}
 registry=https://registry.npmjs.org/
 always-auth=false"
 `;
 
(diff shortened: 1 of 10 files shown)


JimiC:
Suggestion: Since the refactoring done in this PR interferes with #42, I could also incorporate those changes in this PR.

DanielRuf:
`latest` would be great, as a solution you can already use `'*'`.

ExE-Boss:
This PR seems to have a lot of out of scope changes in it.

JimiC:
Not going to update it if it's not going to be merged.

wdanilo:
Why is this not merged? It would be important to me to indicate that I want my tests to run on the latest LTS node version. Right now I need to update it everywhere by hand.

maxim-lobanov:
Fixed in https://github.com/actions/setup-node/pull/270

PR #62, opened 2019-09-18
Fix token references

domoritz:


domoritz:
NM, you use a different registry.

Issue #53, opened 2019-09-04
Unable to publish to GitHub Package Registry


daquinoaldo:
I use the workflow you suggest, but the `npm publish` action on GitHub Package Registry doesn't work. On npmjs it works.

Workflow
```
name: publish

on: release

jobs:
  test:
    name: Test
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v1
      - uses: actions/setup-node@v1
        with:
          node-version: 12
      - run: npm ci
      - run: npm test

  # works
  publish-npm:
    name: Publish NPM
    needs: test
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v1
      - uses: actions/setup-node@v1
        with:
          node-version: 12
          registry-url: https://registry.npmjs.org/
      - run: npm publish
        env:
          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}

  # doesn't work: tries to publish on https://registry.npmjs.org
  publish-gpr:
    name: Publish GitHub Packages
    needs: test
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v1
      - uses: actions/setup-node@v1
        with:
          registry-url: 'https://npm.pkg.github.com'
      - run: npm publish
        env:
          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```

npm publish on npmjs works. On gpr doesn't. Follows the GitHub Action output of the publish-gpr job.

```
Run actions/setup-node@v1
   with:
    registry-url: https://npm.pkg.github.com
    node-version: 10.x
in/tar xzC /home/runner/work/_temp/bc5bc42b-b229-41ac-b8a0-e554467f0c90 -f /home/runner/work/_temp/5cda4e3f-6051-4aaa-b41f-c4761c6cac12
##[add-path]/opt/hostedtoolcache/node/10.16.3/x64/bin
##[set-env name=NPM_CONFIG_USERCONFIG;]/home/runner/work/_temp/.npmrc
##[set-env name=NODE_AUTH_TOKEN;]XXXXX-XXXXX-XXXXX-XXXXX
Added matchers: 'tsc'. Problem matchers scan action output for known warning or error strings and report these inline.
##[add-matcher]/home/runner/work/_actions/actions/setup-node/v1/.github/tsc.json
Added matchers: 'eslint-stylish'. Problem matchers scan action output for known warning or error strings and report these inline.
[...]

Issue #52, opened 2019-09-03
publishing to npm fails after publishing to GPR


getify:
I have a workflow like this that's supposed to publish to npm once I publish to GPR:

```yaml
name: Publish to npm
on: registry_package
jobs:
  publish-npm:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v1
      - uses: actions/setup-node@v1
        with:
          node-version: 12
          registry-url: https://registry.npmjs.org/
      - run: npm install && npm publish
        env:
          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
```

I have the `npm_token` secret added to this repository.

When I just published to GPR, it kicked off this workflow job, but it failed at the last step of publishing to npm. The error was from npm saying:

```cmd
npm ERR! code E401
npm ERR! Unable to authenticate, need: Basic realm="GitHub Package Registry"
```

What does this error mean, and how do I fix it? I don't see anything about setting "basic realm" in the recipes for this `setup-node` action.

damccorm:
One thing I notice off the bat is that your registry url is for npm (`https://registry.npmjs.org/`) not GitHub (`https://npm.pkg.github.com`) which means your auth will be set for npm as well. Could you try changing that and see if it fixes it?

getify:
I am trying to publish from GPR to npm, that's why the URL and token are for npm.

nkzawa:
I also had the same issue, and fixed by creating `~/.npmrc` like:
```yaml
- run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_AUTH_TOKEN }}" > ~/.npmrc
- run: npm publish
```

getify:
According to the examples provided for this repo for publishing to npm, presumably the `with: registry_url: ..` is supposed to be taking the place of a local npmrc, and I would also assume the `NODE_AUTH_TOKEN` environment variable is either something that this action uses or that the npm client itself uses.

The approach of `echo`ing out an npmrc is not only hacky but also seems a bit dangerous given that we already have to have an npmrc in the repo to publish to GPR in the first place, which means effectively this `echo` is overwriting the file just before publish.

damccorm:
Sorry, misunderstood - any chance you can share your package.json file? Also, are you able to publish to npm locally with that package.json file?

getify:
Yes I have been publishing to npm directly (by just commenting out the line in my npmrc that directs the publish to GPR) each time that the github action has failed.

[...]

PR #64, opened 2019-09-25
Fixed auth GPR

valmirphp:
I had the same issues mentioned in the issues:
- #53 
- #52 
- #49

Analyzing the code I realized that the file `/home/runner/work/_temp/.npmrc`  was defective.

- run: cat /home/runner/work/_temp/.npmrc

Output:
```
 //npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}
```

> Fix string replace to concatenate NODE_AUTH_TOKEN

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ typings/
 
 # DynamoDB Local files
 .dynamodb/
+
+# IDEs
+.idea/
--- a/lib/authutil.js
+++ b/lib/authutil.js
@@ -43,7 +43,8 @@ function writeRegistryToFile(registryUrl, fileLocation, alwaysAuth) {
         });
     }
     // Remove http: or https: from front of registry.
-    const authString = registryUrl.replace(/(^\w+:|^)/, '') + ':_authToken=${NODE_AUTH_TOKEN}';
+    const authString = registryUrl.replace(/(^\w+:|^)/, '')
+        + ':_authToken=' + process.env['NODE_AUTH_TOKEN'];
     const registryString = scope
         ? `${scope}:registry=${registryUrl}`
         : `registry=${registryUrl}`;
--- a/lib/authutil.js
+++ b/lib/authutil.js
@@ -43,8 +43,9 @@ function writeRegistryToFile(registryUrl, fileLocation, alwaysAuth) {
         });
     }
     // Remove http: or https: from front of registry.
-    const authString = registryUrl.replace(/(^\w+:|^)/, '')
-        + ':_authToken=' + process.env['NODE_AUTH_TOKEN'];
+    const authString = registryUrl.replace(/(^\w+:|^)/, '') +
+        ':_authToken=' +
+        process.env['NODE_AUTH_TOKEN'];
     const registryString = scope
         ? `${scope}:registry=${registryUrl}`
         : `registry=${registryUrl}`;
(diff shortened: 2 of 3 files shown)


negebauer:
The idea behind the current code, as I understand it, is so that the token isn't saved in the file.
When running commands like `npm install` you can pass the token and it works, as shown in the _Use private packages_ section of the [README](https://github.com/actions/setup-node/blob/master/README.md).

```yaml
- run: npm install
  env:
    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```

valmirphp:
@negebauer  do you run "npm install" with github private packages, following this example?

negebauer:
Yes, and it works.

```yml
    steps:
      - uses: actions/checkout@v1
      - name: Use Node.js ${{ matrix.node-version }}
        uses: actions/setup-node@master
        with:
          node-version: ${{ matrix.node-version }}
          registry-url: https://npm.pkg.github.com
          scope: '@mtblc'
      # From https://github.com/actions/setup-node "Use private packages:"
      # Skip post-install scripts here, as a malicious
      # script could steal NODE_AUTH_TOKEN.
      - run: yarn
        env:
          NODE_AUTH_TOKEN: ${{ secrets.MTBLC_NPM_TOKEN }}
```

Thats the code I'm running

getify:
@negebauer it's broken for many of us (as evidenced in the threads linked by the OP).

getify:
@valmirphp Does npmrc format support a bash-like syntax for pulling the token from the shell environment? If so, wouldn't it be preferable to access it that way?

If not, then it seems like the only way is what you're doing, which is dropping in the NPM token into the auto-generated npmrc.

negebauer:
It does read the token from the shell environment when running `npm install`. You can test that by doing it locally. Modify your `.npmrc` to use an env var, export the env var and `npm install`.

getify:
I think this is actually the problem:

https://github.com/actions/setup-node/blob/master/src/authutil.ts#L57

I think the `NODE_AUTH_TOKEN` environment variable itself is not in fact being exported correctly, not that npmrc isn't picking it up.

It would be useful to verify if that's indeed the case.

negebauer:
That's an empty export on the `setup-node` action. When running `npm install` you should pass `NODE_AUTH_TOKEN` as an env variable. This way the actual token is only available on the install action.

getify:
@negebauer I don't think you're understanding the context of the problems being face. Have you read #52?

negebauer:
Yes I have, but this doesn't fix that. This PR writes a particular env var to `.npmrc` which doesn't need to be written in this step, since it's read from the environment on the steps that actually need it.

valmirphp:
@getify you are absolutely right.

`.npmrc` supports variable ENV, which is the best way to work without physically saving the token.

The PR I created solves this authentication problem, but I believe the way is to solve the export problem.

I will close this PR, and I will look for a new solution.

For now if anyone wants can use my fork, follows example:


```yml
name: Test Setup Node

on: [push]

jobs:
  build:

    runs-on: ubuntu-18.04

    strategy:
      matrix:
        node: [ '10', '8' ]

    steps:
    - uses: actions/checkout@v1
    - name: Use Node.js ${{ matrix.node-version }}
      uses: valmirphp/setup-node@master
      with:
        node-version: ${{ matrix.node-version }}
        registry-url: 'https://npm.pkg.github.com'
        always-auth: false
        scope: '@your-organization'
      env:
        NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN_NPM }}
    - name: Installation
      run: yarn install
      env:
        CI: true
```

PR #71, opened 2019-10-18
Update README to include yarn

ethomson:
A user indicated that it was not clear that the `setup-node` action took care of yarn, and tried using an incompatible third-party action.  Better specify that yarn is included above the fold.


allout58:
Only issue is it won't currently install Yarn (for example, on a self-hosted runner)

bryanmacfarlane:
allout58 is correct.  doesn't install and add yarn to path.  only node/npm

PR #76, opened 2019-10-23
Log npm & yarn versions

Andarist:
This is helpful debug info - IMHO makes sense to always include it.


milesj:
Was just looking for this. Was hoping this action would log info similar to `actions/checkout`.

bryanmacfarlane:
I think we should protect the outputting of the version(s) with io.which call.  e.g. yarn may or may not be installed.  even in the case where a version spec could not be met it may not be installed.  could make into a common function execIfexist(tool, arg).

Babesss:
https://github.com/actions/setup-node/pull/76.patch[](https://github.com/actions/setup-node/pull/76.patch)

Mohsen-Phoenix (approved):

Andarist:
@bryanmacfarlane i've implemented your suggested changes. Could you take another look at this PR?

dmitry-shibanov:
Hello everyone. I'm going to close the pull request in favour of this https://github.com/actions/setup-node/pull/368

PR #95, opened 2019-12-13
Tslint prose matcher

michaelvdnet:
This PR aims to add support for the default tslint formatter: `prose`.

As tslint uses absolute paths, we need to filter part of the path.

It currently assumes the following path:
`/home/runner/work/<repo name>/<repo name>/<path in repo>`

Can anyone confirm that the path will always follow the format above?

### DIFF
--- a/.github/tslint-prose.json
+++ b/.github/tslint-prose.json
@@ -1,0 +1,17 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "tslint-prose",
+      "pattern": [
+        {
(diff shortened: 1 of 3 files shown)


allout58:
I can confirm that it definitely **won't** always be that path, or at least that prefix. I'm working with Self-Hosted runners, so it's completely different (`/home/github/_work` for example).
If the `<repo>/<repo>` part is static, maybe you could have that be the marker for knowing when to start the path group?

Issue #61, opened 2019-09-17
How do I choose the latest node version?


samuelgozi:
I'm sorry if this turns out to be a stupid question, but how do I choose the latest Node version available?
When generating stuff like code coverage(for example), it doesn't matter to me which Node version is used and I rather use the latest one.

fregante:
I was looking for this too but I think there's no way. The version _seems_ to be passed to `semver` for validation and never once _seems_ to refer to tags like `latest`/`stable`/`lts` etc

https://github.com/actions/setup-node/blob/e565252a9dec30354d1b3507ab733e40b9580cc9/lib/installer.js#L53

samuelgozi:
@fregante Thanks for the help. I guess I'll have to set it manually, or just forget abou it.

fregante:
If you use a docker image instead of `setup-node`, I think you can use `node:current` (untested)

```yml
jobs:
  my_job:
    container: node:current
    steps:
    - run: npm install
    - run: npm test
```

samuelgozi:
@fregante Thanks that might help!

DanielRuf:
It would be the `latest` now.

See https://hub.docker.com/_/node?tab=description

![grafik](https://user-images.githubusercontent.com/827205/72882682-50eb1000-3d03-11ea-973b-534c635648e1.png)

DanielRuf:
I checked it and `'*'` as Node version seems to work. Can you verify that?
We should document this and some more examples.

https://www.npmjs.com/package/semver#x-ranges-12x-1x-12-

DanielRuf:
Running the docker container might not be a solution, especially when you have to run arbitrary `docker` commands as part of your tests which will fail then.

PR #104, opened 2020-01-23
Document supported SemVer ranges with example

DanielRuf:
Let's document this. Reading the code does not replace a good documentation.

See https://github.com/actions/setup-node/issues/61

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   <a href="https://github.com/actions/setup-node"><img alt="GitHub Actions status" src="https://github.com/actions/setup-node/workflows/Main%20workflow/badge.svg"></a>
 </p>
 
-This action sets by node environment for use in actions by:
+This action sets the node environment for use in actions by:
 
 - optionally downloading and caching a version of node - npm by version spec and add to PATH
 - registering problem matchers for error output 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,12 @@ This action sets by node environment for use in actions by:
 
 See [action.yml](action.yml)
 
+You can use any valid [SemVer](https://semver.org/) range supported by the [sem-ver](https://www.npmjs.com/package/semver#ranges) module.
+
+For example [`node-version: '*'`](https://www.npmjs.com/package/semver#x-ranges-12x-1x-12-) would satisfy the latest version.
+
+See also the [`evaluateVersions`](https://github.com/actions/setup-node/blob/93313caf82afbdb49b8e4790ac7f1c97c8034826/src/installer.ts#L114) method and its usage of [`semver.satisfies`](https://github.com/actions/setup-node/blob/93313caf82afbdb49b8e4790ac7f1c97c8034826/src/installer.ts#L130)
```
bryanmacfarlane, line 20: would prefer to keep links to code (which goes stale over time) out of the main readme.  this is more of why the feature works.  I think it's enough to just say the preceeding points about any semver is supported.  so would prefer to remove this paragraph

README.md
```diff
@@ -13,6 +13,12 @@ This action sets by node environment for use in actions by:
 
 See [action.yml](action.yml)
 
+You can use any valid [SemVer](https://semver.org/) range supported by the [sem-ver](https://www.npmjs.com/package/semver#ranges) module.
+
+For example [`node-version: '*'`](https://www.npmjs.com/package/semver#x-ranges-12x-1x-12-) would satisfy the latest version.
+
```
bryanmacfarlane, line 19: can point here for more examples: https://github.com/actions/setup-node/blob/main/.github/workflows/versions.yml

bryanmacfarlane:
any interest in updating to target main instead of master?

DanielRuf:
> any interest in updating to target main instead of master?

Done. Normally you should also be able to do that.
Since there are conflicts and this is probably completely out of date, I will close this PR.

PR #105, opened 2020-01-24
follow proxy settings

ericsciple:


.github/workflows/workflow.yml
```diff
@@ -1,25 +1,70 @@
...
+  push:
+    branches:
+      - master
+      - releases/*
+
 jobs:
-  run:
-    name: Run
+  build:
+    name: Build
```
ericsciple: The only change to this job is, i updated the build to use node 12 instead of node 10.

Everything else in this job is just formatting changes... Since I added two more jobs, i changed up the formatting a little to improve skimmability

.github/workflows/workflow.yml
```diff
@@ -1,25 +1,70 @@
...
+        uses: ./
+        with:
+          node-version: 10.x
 
-    - name: npm install
-      run: npm install
+      - name: Verify node version
+        run: __tests__/verify-node-version.sh 10
+
+  test-proxy:
```
ericsciple: proxy e2e

.github/workflows/workflow.yml
```diff
@@ -1,25 +1,70 @@
...
+      options: --dns 127.0.0.1
+    services:
+      squid-proxy:
+        image: datadog/squid:latest
+        ports:
+          - 3128:3128
+    env:
+      https_proxy: http://squid-proxy:3128
+    steps:
+      - uses: actions/checkout@users/ericsciple/m165proxy
```
ericsciple: switch to `@v2` after it's published

lib/installer.js
```diff
@@ -17,10 +17,11 @@ var __importStar = (this && this.__importStar) || function (mod) {
 Object.defineProperty(exports, "__esModule", { value: true });
 // Load tempDirectory before it gets wiped by tool-cache
 let tempDirectory = process.env['RUNNER_TEMPDIRECTORY'] || '';
+const assert = __importStar(require("assert"));
```
ericsciple: ignore this file, review the ts file instead

package.json
```diff
@@ -22,11 +22,11 @@
   "author": "GitHub",
   "license": "MIT",
   "dependencies": {
-    "@actions/core": "^1.0.0",
-    "@actions/github": "^1.0.0",
-    "@actions/io": "^1.0.0",
-    "@actions/tool-cache": "^1.0.0",
-    "typed-rest-client": "^1.5.0",
+    "@actions/core": "^1.2.2",
```
ericsciple: i updated all the toolkit packages

src/installer.ts
```diff
@@ -100,9 +101,17 @@ async function queryLatestMatch(versionSpec: string): Promise<string> {
 
   let versions: string[] = [];
   let dataUrl = 'https://nodejs.org/dist/index.json';
-  let rest: restm.RestClient = new restm.RestClient('setup-node');
-  let nodeVersions: INodeVersion[] =
-    (await rest.get<INodeVersion[]>(dataUrl)).result || [];
+  let httpClient = new hc.HttpClient('setup-node', [], {
```
ericsciple: i added retries here when downloading the versions JSON

further below when the exe is downloaded, `toolCache.downloadTool` already has retries built-in

.github/workflows/workflow.yml
```diff
@@ -1,25 +1,70 @@
...
     steps:
-    - uses: actions/checkout@master
-    
-    - name: Set Node.js 10.x
-      uses: actions/setup-node@master
-      with:
-        version: 10.x
+      - uses: actions/checkout@v2
+      
+      - name: Setup node 12
```
bryanmacfarlane: Thanks.  I had this on my TODO list to update ;)

.github/workflows/workflow.yml
bryanmacfarlane: ... and by that you mean checkout v2.  yep.

bryanmacfarlane:
the code changes LGTM.

However, as I review the changes and see the endless node_modules diffs, I wonder if we just shouldn't do the ncc build as part of this change.  I think we should do the change, point `v1` at the sha in master and get rid of the releases/v1 tag from that alternative approach (but ensure anyone binding to a sha matches)

.github/workflows/workflow.yml
ericsciple: yes

package.json
```diff
@@ -22,29 +24,22 @@
...
     "@types/node": "^12.0.4",
     "@types/semver": "^6.0.0",
-    "husky": "^2.3.0",
+    "@zeit/ncc": "^0.21.0",
     "jest": "^24.8.0",
     "jest-circus": "^24.7.1",
     "prettier": "^1.17.1",
     "ts-jest": "^24.0.2",
     "typescript": "^3.5.1"
-  },
```
bryanmacfarlane: yeah!

bryanmacfarlane:
LGTM - I have some thoughts about pack vs build but let's take as a separate discussion.

package.json
```diff
@@ -3,12 +3,14 @@
   "version": "1.0.0",
```
TingluoHuang: bump version?

ericsciple:
@bryanmacfarlane are you suggesting `npm run build` should `tsc && ncc build`? i like that

ericsciple:
actually i forgot, should format happen before pack. i wonder if `npm run all` should format before build. Then it would be ok to combine pack and build.

PR #110, opened 2020-02-05
consume getJson function for better error messages

ericsciple:
also updated package.json scripts to align with setup-ruby, also ran npm audit fix

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -22,7 +22,6 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm run format-check
-      - run: npm run pack
       - run: npm test
       - name: Verify no unstaged changes
         if: runner.os != 'windows'
--- a/package.json
+++ b/package.json
@@ -5,12 +5,11 @@
   "description": "setup node action",
   "main": "lib/setup-node.js",
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && ncc build",
     "format": "prettier --write **/*.ts",
     "format-check": "prettier --check **/*.ts",
-    "pack": "ncc build",
     "test": "jest",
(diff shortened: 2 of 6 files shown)


.github/workflows/workflow.yml
```diff
@@ -22,7 +22,6 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm run format-check
-      - run: npm run pack
```
ericsciple: `npm run build` now runs ncc

package.json
```diff
@@ -5,12 +5,11 @@
   "description": "setup node action",
   "main": "lib/setup-node.js",
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && ncc build",
```
ericsciple: aligned with the pattern set by setup-ruby

src/installer.ts
```diff
@@ -105,13 +105,8 @@ async function queryLatestMatch(versionSpec: string): Promise<string> {
...
     maxRetries: 3
   });
-  let response = await httpClient.get(dataUrl);
-  assert.ok(
-    response.message.statusCode === 200,
-    `Unexpected HTTP status code '${response.message.statusCode}'`
-  );
-  let body = await response.readBody();
-  let nodeVersions = JSON.parse(body) as INodeVersion[];
+  let response = await httpClient.getJson<INodeVersion[]>(dataUrl);
```
ericsciple: getJson now checks for errors, so i removed the error checking code here

src/installer.ts
```diff
@@ -105,13 +105,8 @@ async function queryLatestMatch(versionSpec: string): Promise<string> {
...
+  let nodeVersions = response.result || [];
```
ericsciple: `|| []` is to satisfy typescript. Shouldnt be null since getJson checks for errors

Issue #87, opened 2019-11-16
.npmrc file with registry= not overwritten and causing 401 error


rynz:
Example: https://github.com/rynz/node-test-protocol/pull/6

In this example exists an empty `.npmrc` and when it is populated with `registry=` an error occurs:
```
npm ERR! 401 Unauthorized - GET https://npm.pkg.github.com/rynz/prettier - could not retrieve user by token
```

For whatever reason, the `registry=` is not being removed correctly like https://github.com/actions/setup-node/blob/master/src/authutil.ts#L41 is supposed to do.

joebowbeer:
Your scoped registry should be defined as:

`@rynz/registry=https://npm.pkg.github.com/`

I think it is a bug that setup-node tries to remove `registry`. That is the default registry for non-scoped dependencies, which should be unrelated to your private packages.

I think setup-node should only be removing the lines that it is redefining.

rynz:
Interesting, I'm not sure why I used the scoped registry style I did as it was a long time ago.
Should it be defined as:

`@rynz:registry=https://npm.pkg.github.com/`

Why did you choose to use `/`?

joebowbeer:
The / doesn't seem to matter, at least for the GitHub registry, but it seems to be standard so I added it.

`@rynz:registry` defines where `@rynz` packages are located. `registry` defines the default registry for unscoped (and possibly unresolved?) dependencies.

The GPR documentation has been inconsistent and sometimes misleading but I've cleared away some confusion by looking at npm and setup-node source.

rynz:
Yes I believe it was from the documentation that I used the scoped style I did. From the latest npm documentation it looks to be `@myscope:registry=https://mycustomregistry.example.org` is the standard now. Thanks for pointing me in the right direction @joebowbeer appreciate it.

rynz:
Reopening because I've found the documentation that causes the bug in the first place: https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-npm-for-use-with-github-packages#publishing-a-package-using-a-local-npmrc-file

It states to use, `registry=https://npm.pkg.github.com/OWNER` where as we should be using `OWNER:registry=https://npm.pkg.github.com/`. Perhaps the documentation should be updated?

joebowbeer:
> It states to use, `registry=https://npm.pkg.github.com/OWNER` where as we should be using `OWNER:registry=https://npm.pkg.github.com/`. Perhaps the documentation should be updated?

[...]

Issue #77, opened 2019-10-23
Downloading from nodejs.org failing


paglias:
I'm not sure this is the best repo to report the issue, if it's the wrong place please close this :)

We've been using Github Actions successfully for a few days but starting today all our builds have started to fail at the `setup-node` step with this error:

```

4m 57s
##[error]Node run failed with exit code 1
Run actions/setup-node@v1
  with:
    node-version: 12.x
    always-auth: false
/bin/tar xzC /home/runner/work/_temp/83341e1a-c86d-4ddb-9963-ef01607f6e1e -f /home/runner/work/_temp/58fc8365-fe95-474e-ad3c-c5511594abdd

gzip: stdin: unexpected end of file
/bin/tar: Unexpected EOF in archive
/bin/tar: Unexpected EOF in archive
/bin/tar: Error is not recoverable: exiting now
##[error]The process '/bin/tar' failed with exit code 2
##[error]Node run failed with exit code 1
```

Attaching a screenshot for more clarity
<img width="1114" alt="Schermata 2019-10-23 alle 16 22 25" src="https://user-images.githubusercontent.com/589911/67403285-faf6b700-f5b1-11e9-98fe-fa0c96a59aec.png">

You can see an example of a failed build here https://github.com/HabitRPG/habitica/pull/11474/checks?check_run_id=271716057.

The code of the workflow is the standard one that can be seen in the official docs

```
steps:
.....
    - name: Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v1
      with:
        node-version: ${{ matrix.node-version }}
....
```

rashedmyt:
I am also having the same issue

chingc:
I'm also having the same problem.  One thing I noticed on my CircleCI builds is nvm is failing to download nodejs.  The download is so slow that it will timeout, then the installation fails due to an incomplete download.

I'm think it might be a similar problem affecting this action.  If it's the same issue, it would be nice if there was some kind of automatic resume so it doesn't keep trying to download from the beginning.

abatilo:
I'm having the same problem!

dijam:
I retied couple of times. Waited 10 min, a cup of coffee, tried again and it worked.

vibou:
Same here ! What's going on !

timofeee:
Just experimenting right now with Actions, good to know it's not something I messed up

nicubarbaros:
Same here! AAF

andrew-boyd:
[...]

PR #21, opened 2019-08-05
Add auth support

damccorm:

PR #117, opened 2020-02-28
nodejs.yml

yanz-androidify:
#21 #87 #77


yanz-androidify:
#117

okuryu:
We have already a workflow.
https://github.com/actions/setup-node/blob/master/.github/workflows/workflow.yml

PR #123, opened 2020-03-21
adjust a few examples to mention nodejs 12

staabm:
make users aware that nodejs 12 is also supported


okuryu:
We have already a similar proposal: #108

Issue #132, opened 2020-04-06
Action timing out when new node versions are released


ebwinters:
I am using this in a job:
```yaml
name: Use Node.js ${{ matrix.node-version }}
uses: actions/setup-node@v1
  with:
    node-version: ${{ matrix.node-version }}
```
and `node-version` `13.x`, but my workflow takes 10+minutes just to complete use node. Any idea why or anyone else having this problem?

Xetera:
Same here, It looks like I'm sometimes timing out and getting
> Premature close
> Waiting 16 seconds before trying again
> Premature close
> Waiting 18 seconds before trying again
> ##[error]Premature close

and sometimes it's working fine but still taking 7-11 minutes per install for some reason (12.x)

ebwinters:
good to know other people are having issues - seriously stalling my pipeline 😞

camelmasa:
Same here, We are using node 12.x.

fishcharlie:
Same. You can see the workflow on an open source project experiencing this [here](https://github.com/dynamoosejs/dynamoose/pull/779/checks?check_run_id=563127394). From my experience this just started happening recently.

As you can see the setup node step took 7 minutes and 34 seconds 😱 

<img width="1106" alt="image" src="https://user-images.githubusercontent.com/860375/78520227-0fef8d00-7783-11ea-9db6-72201b933f7d.png">

fishcharlie:
One of the **major** issues of this action is the fact that there is next to no logs that get printed. I brought this up in #90 for the version it's using. But this is another situation where **we need logs**!!

jiyuhan:
This is a serious issue if one's pipeline doesn't have default timeout setup. This will cause a lot of people's CI to hang for up to 6 hours. This is not wasting MS' resource, but is wasting resource for other paying customers. Please fix this!

boredland:
this did cost us a lot of money already. please have a look!

matiaschileno:
We have the same problem. Today every pipeline is failing because of this. We are using Node.js 12.x.

ssannidhi-palo:
Experiencing similar issues with Node 12.x version. The step is either timing out or takes up too much time. (~25 mins)

![image](https://user-images.githubusercontent.com/54425886/78541442-374c6580-7828-11ea-81e8-0ff0734bd91e.png)

robbruce:
Cloned the repo and ran the unit tests, it would appear that the issue is with nodejs.org

```
[...]

PR #134, opened 2020-04-06
Added `mirror` option to select alternative download source

### LINKED ISSUE #65: Support specifying alternate node mirrors
I have a use case where I test against latest node LTS, latest node stable *and* latest [node on V8 lkgr](https://github.com/nodejs/node-v8), with the latter requiring to use a different than the hardcoded [dataUrl](https://github.com/actions/setup-node/blob/e565252a9dec30354d1b3507ab733e40b9580cc9/src/installer.ts#L102) ([this one](https://nodejs.org/download/v8-canary/index.json) to be precise).

Noticed that the latest tags are being in the works already, which is great, so I thought I mention the mirror as well since that would allow me to replace my custom and hopefully temporary [solution](https://github.com/dcodeIO/setup-node-nvm) with proper setup-node that works on all platforms :)

daniellockyer:
Fixes #65.

A lot of us are getting burned by the ([ongoing](https://github.com/nodejs/node/issues/32683)) intermittent outage at Node which is [stopping](https://github.com/actions/setup-node/issues/132) `setup-node` from downloading the binaries. It'd be nice to be able to use a mirror where appropriate.

This PR fixes an old issue on the repo to implement support for using mirrors.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -8,6 +8,9 @@ inputs:
   node-version:
     description: 'Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0'
     default: '10.x'
+  mirror:
+    description: 'Optional mirror to download binaries from'
+    default: 'https://nodejs.org/dist/'
   registry-url:
     description: 'Optional registry to set up for auth. Will set the registry in a project level .npmrc and .yarnrc file, and set up auth to read in from env.NODE_AUTH_TOKEN'
   scope:
--- a/dist/index.js
+++ b/dist/index.js
@@ -4487,7 +4487,7 @@ module.exports = require("https");
 /***/ 215:
 /***/ (function(module) {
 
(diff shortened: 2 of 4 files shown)


chingc:
Is there a list of official mirrors?  It would be great if it would automatically retry with a mirror upon failure.

daniellockyer:
@chingc Not that I can see. There are a few mirrors I can find from a quick Google but none officially recommended.

dcodeIO:
Not sure if relevant, but one mirror I'm using is https://nodejs.org/download/v8-canary/ for WebAssembly testing, which provides versions of node built against v8 LKGR. Having support for that in setup-node would be great.

thomas-z-meyer-db:
@daniellockyer hi, why way this closed? mirror support would be really nice, especially in more involved setups.

zhyc9de:
I really need this feature, too

Issue #90, opened 2019-11-23
Log Node.js version using


fishcharlie:
Couple of things.

First if you specify `node-version` of `10`, does that use the absoute latest version of Node.js v10 (both minor and patch versions)?

Second, does `node-version` of `10.x` use the latest version when taking account patch versions? Or does `.x` just mean the minor version?

Finally, it'd be great if this action printed the Node.js version it decided to use in the logs of GitHub Actions so developers can determine which Node.js version it decided to use, which would make debugging SO much easier.

<img width="1055" alt="Screen Shot 2019-11-22 at 3 49 46 PM" src="https://user-images.githubusercontent.com/860375/69465889-bb052a00-0d3f-11ea-8f40-41f5e163aecf.png">

joshmgross:
👋 Hey @fishcharlie. Specifying a `node-version` of `10` and `10.x` will have the same result. This follows [semver](https://semver.org/), where `10 == 10.x == 10.x.x`, which is effectively "the latest version of Node.js v10 taking into account both minor and patch versions. 

>Finally, it'd be great if this action printed the Node.js version it decided to use in the logs of GitHub Actions

Great idea! That should be a quick fix.

bryanmacfarlane:
Yes, it should work with any semver version spec.  As part of this issue, I'll make sure it's logged as well.  I'll also update the readme / docs

fishcharlie:
Any updates on this? cc/ @bryanmacfarlane @joshmgross

joshmgross:
@bryanmacfarlane are you still planning on doing this? If not,  I can pick this up.

bryanmacfarlane:
@joshmgross - I'm fine with you picking this up.  If you do, look at recent commits in setup-go where I did it.  I think it should also output npm version.

fishcharlie:
Outputting the npm version too would be amazing!!!

demian85:
Can someone please explain why my node version is not 12 as specified?

Part of my config file is
```
jobs:
  primary:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        node-version: [12.x]

    steps:
      - uses: actions/checkout@v1

      - name: Cache node modules
        uses: actions/cache@v1.1.0
        with:
          path: ~/.npm
          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
          restore-keys: |
            ${{ runner.os }}-node-
      - name: Node ${{ matrix.node-version }}
[...]

PR #137, opened 2020-04-16
Log node and npm versions

joshmgross:
Resolves https://github.com/actions/setup-node/issues/90

This logs the `node` and `npm` versions that are installed by `setup-node`, following the pattern established by `setup-go`, see https://github.com/actions/setup-go/blob/v2-beta/src/main.ts#L54

### Example
**Run:** https://github.com/actions/setup-node/pull/137/checks?check_run_id=606120085#step:4:5
**Log output**
```
C:\hostedtoolcache\windows\node\0.12.18\x64\node.exe --version
v0.12.18
C:\windows\system32\cmd.exe /D /S /C ""C:\Program Files\nodejs\npm.cmd" --version"
6.14.4
```


src/setup-node.ts
```diff
@@ -17,6 +19,15 @@ async function run() {
       await installer.getNode(version);
     }
 
+    // Output version of node and npm that are being used
+    const nodePath = await io.which('node');
+    const nodeVersion = cp.execSync(`${nodePath} --version`);
```
bryanmacfarlane, line 24: Doesn't execSync return a buffer of bytes that you need toString() (safely)?  See https://github.com/actions/setup-go/blob/master/src/main.ts#L56

joshmgross: I thought template strings (ex `${nodePath} --version`) already called `.toString()`

bryanmacfarlane: Ahhh - yeah.  I think you're right.  Just e2e test

src/setup-node.ts
```diff
@@ -17,6 +19,15 @@ async function run() {
...
+    console.log(`Node Version: ${nodeVersion}`);
+
+    const npmPath = await io.which('npm');
+    const npmVersion = cp.execSync(`${npmPath} --version`);
+    console.log(`npm Version: ${npmVersion}`);
```
bryanmacfarlane, line 29: not sure why none of this would fail.  On one hand I was thinking we should make all this best effort.  But then I rethought that if this can't work, then the action didn't do it's job so it should fail.  So I think it's fine how it is.

fishcharlie:
image

Looks like this new code might be failing on Windows??

bryanmacfarlane:
@joshmgross ☝️

joshmgross:
Thanks @fishcharlie, it was an issue with the spaces in the Windows Path, causing the path to not be interpreted correctly by `cp.execSync`

I've switched to `@actions/exec` which handles this edge case.

fishcharlie:
@joshmgross Perfect. Looks great! Thanks for this. Never seen `"$ --version` syntax before (`"$` part). But I'm assuming it works fine. Would have suggested writing some tests just to verify, but not sure how you would really write tests cases around this.

Looking forward to seeing this be merged and released as part of this action. Thanks again!!

joshmgross:
>Never seen "$ --version syntax before ("$ part).

Bug on my part, but should be fixed now. 

>Would have suggested writing some tests just to verify, but not sure how you would really write tests cases around this.

I think we would have to mock out `exec` to get any testing done, and at that point, I'm not sure what we would actually be testing. I'm letting the E2E test cover this case.

src/setup-node.ts
```diff
@@ -1,4 +1,6 @@
 import * as core from '@actions/core';
+import * as exec from '@actions/exec';
+import * as io from '@actions/io';
```
XhmikosR, line 3: Isn't this unused @joshmgross?

PR #142, opened 2020-04-21
Run test job on windows-latest

joshmgross:
This adds running `windows-latest` to the set of OS's that we run E2E tests on. This ensures we have coverage for non-nix based operating systems.


bryanmacfarlane:
LGTM

PR #147, opened 2020-04-24
download from node-versions and fallback to node dist

bryanmacfarlane:
Node distributions are now available from https://github.com/actions/node-versions for reliability since the releases are backed by a CDN.  That repo offers a versions-manifest.json which abstracts away the location of each version.  Therefore in the future it could move from releases to some other CDN or endpoint.  This set of changes queries the versions-manifest against the node version semver spec provided by the workflow to resolved to latest for that spec.  If not found in the versions-manifest, it falls back to the node dist ( https://nodejs.org/dist/index.json )


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,32 @@
+name: build-test
```
bryanmacfarlane, line 1: I separated these into separate workflows so I can show badges for build-test, versions and proxy status on the readme.  In general I like the separation.

.gitignore
```diff
@@ -2,6 +2,9 @@ node_modules/
 lib/
 __tests__/runner/*
 
+validate/temp
+validate/node
```
bryanmacfarlane, line 6: I created a bash script to automate interactively validating without queuing a full run

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug Jest Tests on Nix",
```
bryanmacfarlane, line 8: I was able to step through jest tests with this.  It works on mac / linux and I saw mention of different configs for windows but I didn't do that yet or validate so I called Nix.

__tests__/authutil.test.ts
```diff
@@ -1,67 +1,123 @@
...
+    process.stderr.write('dbg::' + message + '::\n');
+  }
+
+  afterAll(async () => {
+    if (_runnerDir) {
+      await io.rmRF(_runnerDir);
+    }
+  }, 100000);
+
+  function readRcFile(rcFile: string) {
```
bryanmacfarlane, line 66: when I upgraded to the new typescript compiler and jest version the snapshot stuff was providing more problems than value.  I got it working by updating the baselines then after changes and updating it stopped working on CI even though I validated with debug code that it was equivalent byte for byte.  It was saying I need to update baselines on CI server even though I did.  In the end, comparing a couple keys programmatically was much simpler.

__tests__/data/versions-manifest.json
```diff
@@ -0,0 +1,152 @@
+[
```
bryanmacfarlane, line 1: this is a minimal version of versions-manifest @ https://github.com/actions/node-versions/ - just enough versions to unit test resolution

action.yml
```diff
@@ -1,17 +1,21 @@
...
-    default: '12.x'
+    description: 'Version Spec of the version to use.  Examples: 12.x, 10.15.1, >=10.15.0'
   registry-url:
     description: 'Optional registry to set up for auth. Will set the registry in a project level .npmrc and .yarnrc file, and set up auth to read in from env.NODE_AUTH_TOKEN'
   scope:
     description: 'Optional scope for authenticating against scoped registries'
+  token:
+    description: Used to pull node distributions from node-versions.  Since there's a default, this is typically not supplied by the user.
+    default: ${{ github.token }}
+# TODO: add input to control forcing to pull from cloud or dist.
```
bryanmacfarlane, line 17: I'm going to do this as a separate PR while in v2-beta

action.yml
```diff
@@ -1,17 +1,21 @@
...
-    description: 'Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0'
-    default: '12.x'
+    description: 'Version Spec of the version to use.  Examples: 12.x, 10.15.1, >=10.15.0'
   registry-url:
     description: 'Optional registry to set up for auth. Will set the registry in a project level .npmrc and .yarnrc file, and set up auth to read in from env.NODE_AUTH_TOKEN'
   scope:
     description: 'Optional scope for authenticating against scoped registries'
+  token:
+    description: Used to pull node distributions from node-versions.  Since there's a default, this is typically not supplied by the user.
+    default: ${{ github.token }}
```
bryanmacfarlane, line 16: We need to make sure that querying api.github.com for the manifest and release data doesn't get throttled.  Authenticated is per user and much higher (similar to our runner pulling actions) while anon would be throttled much lower and per IP

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "setup-node",
-  "version": "1.0.0",
+  "version": "2.0.0",
```
bryanmacfarlane, line 3: doesn't influence anything but conveys the lineage.

package.json
```diff
@@ -28,18 +28,18 @@
     "@actions/github": "^1.1.0",
     "@actions/http-client": "^1.0.6",
     "@actions/io": "^1.0.2",
-    "@actions/tool-cache": "^1.3.3",
+    "@actions/tool-cache": "^1.5.3",
```
bryanmacfarlane, line 31: This is in preview released from a branch.  As setup-python and setup-node finishes it's changes, we can feel confident in the api changes.

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
 // see https://nodejs.org/dist/index.json
 //
-interface INodeVersion {
+export interface INodeVersion {
   version: string;
   files: string[];
 }
 
-export async function getNode(versionSpec: string) {
+interface INodeVersionInfo {
```
bryanmacfarlane, line 21: These interfaces normalize the resolution from manifest and fallback to node dist so common code downstream can be agnostic.

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
+  fileName: string;
+}
+
+export async function getNode(
+  versionSpec: string,
+  stable: boolean,
+  token: string
+) {
+  let osPlat: string = os.platform();
+  let osArch: string = translateArchToDistUrl(os.arch());
```
bryanmacfarlane, line 34: had to move from module global to function for mocking to work

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
-      // query nodejs.org for a matching version
-      version = await queryLatestMatch(versionSpec);
-      if (!version) {
-        throw new Error(
-          `Unable to find Node version '${versionSpec}' for platform ${osPlat} and architecture ${osArch}.`
-        );
+  if (toolPath) {
+    console.log(`Found in cache @ ${toolPath}`);
+  } else {
+    console.log(`Attempting to download ${versionSpec}...`);
```
bryanmacfarlane, line 45: I added much more output to the logs.  The previous impl was pretty much sllent

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
+    let info = await getInfoFromManifest(versionSpec, stable, token);
+    if (!info) {
+      console.log(
+        'Not found in manifest.  Falling back to download directly from Node'
```
bryanmacfarlane, line 49: LTS lineages (8,10,12,14) will come from manifest while non-LTS/old versions will still fall through to node dist.  We can choose to pull more into manifest if needed.

bryanmacfarlane: compute is also going to do separate changes so the latest of each LTS is pre-cached on the image as part of image gen

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
+    }
+
+    console.log(`Acquiring ${info.resolvedVersion} from ${info.downloadUrl}`);
+
+    let downloadPath = '';
+    try {
+      downloadPath = await tc.downloadTool(info.downloadUrl, undefined, token);
+    } catch (err) {
+      if (err instanceof tc.HTTPError && err.httpStatusCode == 404) {
+        await acquireNodeFromFallbackLocation(info.resolvedVersion);
```
bryanmacfarlane, line 67: a few really old versions like 0.12.8 will have a different layout without a targz or 7z.  it will have just a node binary

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
+    //
+    // Extract
+    //
+    console.log('Extracting ...');
+    let extPath: string;
+    if (osPlat == 'win32') {
+      let _7zPath = path.join(__dirname, '..', 'externals', '7zr.exe');
+      extPath = await tc.extract7z(downloadPath, undefined, _7zPath);
+      // 7z extracts to folder matching file name
+      let nestedPath = path.join(extPath, path.basename(info.fileName, '.7z'));
```
bryanmacfarlane, line 83: node dist creates a 7z with a nested folder matching the filename in the unzipped structure.  The manifest extracts without that.  That's a good thing because later I'm considering extracting directly into the cache instead of unzip and copy.  Copy on windows consumes the vast majority of the time.

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
+      if (fs.existsSync(nestedPath)) {
+        extPath = nestedPath;
+      }
+    } else {
+      extPath = await tc.extractTar(downloadPath, undefined, [
+        'xz',
+        '--strip',
+        '1'
```
bryanmacfarlane, line 91: same here.  strip 1 will strip the root folder if it exists.  I validated it won't if it doesn't.  Note that I extended the interface to accept string | string[] which should be fine for compat.

src/installer.ts
```diff
@@ -65,41 +114,63 @@ export async function getNode(versionSpec: string) {
...
+  versionSpec: string,
+  stable: boolean,
+  token: string
+): Promise<INodeVersionInfo | null> {
+  let info: INodeVersionInfo | null = null;
+  const releases = await tc.getManifestFromRepo(
+    'actions',
+    'node-versions',
+    token,
+    'update-versions-manifest-file' // TODO: remove after testing
```
bryanmacfarlane, line 127: I will remove after Maxim merges his last set of changes today.

src/installer.ts
```diff
@@ -65,41 +114,63 @@ export async function getNode(versionSpec: string) {
...
+
+  let info: INodeVersionInfo | null = null;
+  let version: string;
+
+  version = await queryDistForMatch(versionSpec);
+  if (!version) {
+    return null;
+  }
+
+  //
```
bryanmacfarlane, line 157: This is all mostly the same code - just moved into another function for fallback.

src/main.ts
```diff
@@ -0,0 +1,41 @@
+import * as core from '@actions/core';
+import * as installer from './installer';
+import * as auth from './authutil';
+import * as path from 'path';
+
```
bryanmacfarlane, line 5: moved logic into a main for testing.  Mostly the same logic.

src/main.ts
```diff
@@ -0,0 +1,41 @@
...
+export async function run() {
+  try {
+    //
+    // Version is optional.  If supplied, install / use from the tool cache
+    // If not supplied then task is still used to setup proxy, auth, etc...
+    //
+    let version = core.getInput('node-version');
+    if (!version) {
+      version = core.getInput('version');
```
bryanmacfarlane, line 14: even though this is officially deprecated, I left it in because I wasn't in the mood to break folks as part of these larger changes.

__tests__/authutil.test.ts
```diff
@@ -1,67 +1,123 @@
...
     rcFile = path.join(tempDir, '.npmrc');
   }, 100000);
 
-  beforeEach(() => {
-    if (fs.existsSync(rcFile)) {
-      fs.unlinkSync(rcFile);
-    }
+  beforeEach(async () => {
+    await io.rmRF(rcFile);
+    // if (fs.existsSync(rcFile)) {
```
ericsciple, line 33: dead code

src/installer.ts
```diff
@@ -65,41 +114,62 @@ export async function getNode(versionSpec: string) {
...
+  // Download - a tool installer intimately knows how to get the tool (and construct urls)
+  //
+  version = semver.clean(version) || '';
+  let fileName: string =
+    osPlat == 'win32'
+      ? `node-v${version}-win-${osArch}`
```
joshmgross, line 162: If `version` is an empty string, won't this file name be invalid?

joshmgross (approved):
LGTM. Thanks for all the review comments, makes it a lot easier to get context on the changes

__tests__/__snapshots__/authutil.test.ts.snap
```diff
@@ -1,31 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ericsciple, line 1: i never understood the snapshot thing. Nice to see it deleted :) one less thing to learn

__tests__/authutil.test.ts
ericsciple: +1 i'm honestly glad to see it deleted - one less thing to grok and i never fully understood how it worked.

src/installer.ts
ericsciple: It looks like unzip + copy may be related to the extra folder after unzipping.

Move is unsafe on Windows because defender commonly still has a lock reading the newly created files on disk. Copy is safe.

src/main.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    console.log(`version: ${version}`);
+    if (version) {
+      let token = core.getInput('token');
```
ericsciple, line 19: Consider validating token is supplied. Marketplace template will be:

```yml
uses: actions/node@v2-beta
with:
  token: '' # The auth token
```

In checkout, I have seen customers with `token: ''` i think it's related to the marketplace copy/paste

src/main.ts
```diff
@@ -0,0 +1,41 @@
...
+      let stable = (core.getInput('stable') || 'true').toUpperCase() === 'TRUE';
+      await installer.getNode(version, stable, token);
+    }
+
+    const registryUrl: string = core.getInput('registry-url');
+    const alwaysAuth: string = core.getInput('always-auth');
+    if (registryUrl) {
+      auth.configAuthentication(registryUrl, alwaysAuth);
```
ericsciple, line 27: should this be `await auth.configAuthentication`? The implementation looks "sync" but unit tests await

src/installer.ts
```diff
@@ -65,41 +114,62 @@ export async function getNode(versionSpec: string) {
...
-      break;
-    default:
-      throw new Error(`Unexpected OS '${osPlat}'`);
+async function getInfoFromManifest(
+  versionSpec: string,
+  stable: boolean,
+  token: string
+): Promise<INodeVersionInfo | null> {
+  let info: INodeVersionInfo | null = null;
+  const releases = await tc.getManifestFromRepo(
```
ericsciple, line 123: where is the manifest file?

src/installer.ts
```diff
@@ -1,54 +1,103 @@
...
+      throw new Error(
+        `Unable to find Node version '${versionSpec}' for platform ${osPlat} and architecture ${osArch}.`
+      );
+    }
+
+    console.log(`Acquiring ${info.resolvedVersion} from ${info.downloadUrl}`);
+
+    let downloadPath = '';
+    try {
+      downloadPath = await tc.downloadTool(info.downloadUrl, undefined, token);
```
ericsciple, line 64: shouldnt this use info.token?

ericsciple: when downloading from nodejs.org, shouldnt use a token

ericsciple: or would it better to check the hostname of the download url to determine whether to pass token?

ericsciple (approved):

Thomasdr06:
https://github.com/actions/setup-node/pull/147

package.json
Thomasdr06: https://github.com/actions/setup-node/pull/147#pullrequestreview-404993690

Thomasdr06: https://github.com/actions/setup-node/pull/147#pullrequestreview-404993690

Thomasdr06: https://github.com/actions/setup-node/pull/147#pullrequestreview-404993690

Thomasdr06: https://github.com/actions/setup-node/pull/147#pullrequestreview-404993690}https://github.com/Thomasdr06/setup-node/blob/v-mazhuk/automate-releasing-new-versihttps://github.com/Thomasdr06/setup-node/commit/78148dae5052c4942d5b0f92719061df122a3b1cons/tsconfig.json

PR #157, opened 2020-05-15
ghes support

ericsciple:
1) Remote auth token when calling Dotcom from GHES
2) Make download from Releases best effort

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -1338,7 +1338,7 @@ var __importStar = (this && this.__importStar) || function (mod) {
 };
 Object.defineProperty(exports, "__esModule", { value: true });
 const semver = __importStar(__webpack_require__(280));
-const core_1 = __webpack_require__(470);
+const core_1 = __webpack_require__(902);
 // needs to be require for core node modules to be mocked
 /* eslint @typescript-eslint/no-require-imports: 0 */
 const os = __webpack_require__(87);
@@ -4631,6 +4631,7 @@ const core = __importStar(__webpack_require__(470));
 const installer = __importStar(__webpack_require__(749));
 const auth = __importStar(__webpack_require__(202));
 const path = __importStar(__webpack_require__(622));
+const url_1 = __webpack_require__(835);
 function run() {
     return __awaiter(this, void 0, void 0, function* () {
         try {
@@ -4645,8 +4646,9 @@ function run() {
             console.log(`version: ${version}`);
             if (version) {
                 let token = core.getInput('token');
+                let auth = !token || isGhes() ? undefined : `token ${token}`;
                 let stable = (core.getInput('stable') || 'true').toUpperCase() === 'TRUE';
-                yield installer.getNode(version, stable, token);
+                yield installer.getNode(version, stable, auth);
             }
             const registryUrl = core.getInput('registry-url');
             const alwaysAuth = core.getInput('always-auth');
@@ -4664,6 +4666,10 @@ function run() {
     });
 }
 exports.run = run;
+function isGhes() {
+    const ghUrl = new url_1.URL(process.env['GITHUB_SERVER_URL'] || 'https://github.com');
+    return ghUrl.hostname.toUpperCase() !== 'GITHUB.COM';
+}
 //# sourceMappingURL=main.js.map
 
 /***/ }),
@@ -10883,7 +10889,7 @@ var __importDefault = (this && this.__importDefault) || function (mod) {
     return (mod && mod.__esModule) ? mod : { "default": mod };
 };
 Object.defineProperty(exports, "__esModule", { value: true });
-const core = __importStar(__webpack_require__(470));
+const core = __importStar(__webpack_require__(902));
 const io = __importStar(__webpack_require__(1));
 const fs = __importStar(__webpack_require__(747));
 const mm = __importStar(__webpack_require__(31));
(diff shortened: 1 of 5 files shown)


dist/index.js
```diff
@@ -4618,6 +4618,7 @@ const core = __importStar(__webpack_require__(470));
 const installer = __importStar(__webpack_require__(749));
 const auth = __importStar(__webpack_require__(202));
 const path = __importStar(__webpack_require__(622));
+const url_1 = __webpack_require__(835);
```
ericsciple, line 4621: ignore index.js, review .ts

src/installer.ts
```diff
@@ -42,31 +39,62 @@ export async function getNode(
...
-      );
-    }
-
-    console.log(`Acquiring ${info.resolvedVersion} from ${info.downloadUrl}`);
-
     let downloadPath = '';
+    let info: INodeVersionInfo | null = null;
+
+    //
+    // Try download from internal distribution (popular versions only)
```
ericsciple, line 46: get info + download inside try/catch

The formatting rules bloats the line count quite a bit. We might be on old formatting rules because lots of extra newlines added by formatting rules.

src/installer.ts
```diff
@@ -42,31 +39,62 @@ export async function getNode(
...
+        );
+      } else {
+        console.log(
+          'Not found in manifest.  Falling back to download directly from Node'
+        );
+      }
     } catch (err) {
-      if (err instanceof tc.HTTPError && err.httpStatusCode == 404) {
-        return await acquireNodeFromFallbackLocation(info.resolvedVersion);
+      // Rate limited?
```
ericsciple, line 65: 403 indicates rate limited. Either way print to Info and fallback.

src/installer.ts
```diff
@@ -42,31 +39,62 @@ export async function getNode(
...
+      if (err instanceof tc.HTTPError && err.httpStatusCode === 403) {
+        console.log(
+          'Received HTTP status code 403.  This usually indicates the rate limit has been exceeded'
+        );
+      } else {
+        console.log(err.message);
       }
+      core.debug(err.stack);
```
ericsciple, line 73: I recently started liking err.stack (prints the error message + stack trace). I dont mind yanking this if you prefer. Lmk

src/installer.ts
```diff
@@ -83,6 +111,7 @@ export async function getNode(
     //
     // Install into the local tool cache - node extracts with a root folder that matches the fileName downloaded
     //
+    info = info || ({} as INodeVersionInfo); // satisfy compiler
```
ericsciple, line 114: :sigh: compiler couldnt figure out it's not null by the time it gets here

src/installer.ts
```diff
@@ -103,13 +132,13 @@ export async function getNode(
...
   stable: boolean,
-  token: string
+  token: string | undefined
 ): Promise<INodeVersionInfo | null> {
   let info: INodeVersionInfo | null = null;
   const releases = await tc.getManifestFromRepo(
     'actions',
     'node-versions',
-    token
+    token || ''
```
ericsciple, line 141: Need to modify tc.getManifestFromRepo to allow undefined/empty token.

Today it assumes token is not empty and always adds authorization header.

For E2E testing, i hacked it locally

src/installer.ts
bryanmacfarlane: It's fine.

src/installer.ts
bryanmacfarlane: I thought a 429 would be rate limiting but I guess that's usually - hey back off where 403 is you hit the limit.  I guess a bad token would be 401

src/installer.ts
```diff
@@ -43,39 +40,70 @@ export async function getNode(
...
+        throw err;
+      }
     }
 
     //
     // Extract
     //
     console.log('Extracting ...');
     let extPath: string;
+    info = info || ({} as INodeVersionInfo); // satisfy compiler, never null when reaches here
```
bryanmacfarlane, line 106: cool

src/installer.ts
bryanmacfarlane: OK.  Update it and I'll release today.

src/main.ts
```diff
@@ -39,3 +40,8 @@ export async function run() {
     core.setFailed(error.message);
   }
 }
+
+function isGhes(): boolean {
```
bryanmacfarlane, line 44: might not be a bad idea to add this to toolkit core at some point.

dist/index.js
```diff
@@ -4664,6 +4666,10 @@ function run() {
     });
 }
 exports.run = run;
+function isGhes() {
+    const ghUrl = new url_1.URL(process.env['GITHUB_URL'] || 'https://github.com');
```
ericsciple, line 4670: switch to GITHUB_SERVER_URL

src/main.ts
```diff
@@ -39,3 +41,8 @@ export async function run() {
...
+  const ghUrl = new URL(process.env['GITHUB_URL'] || 'https://github.com');
```
ericsciple, line 46: switch to GITHUB_SERVER_URL

src/main.ts
ericsciple: +1

lets finish rolling out the feature flags first before adding to the toolkit. Ran into a snag after Ring 1 and in the process of switching to GITHUB_SERVER_URL instead of GITHUB_URL

ericsciple:
Wait for https://github.com/actions/toolkit/pull/460 to merge then bump tool-cache

src/installer.ts
ericsciple: Here's the PR https://github.com/actions/toolkit/pull/460

src/installer.ts
ericsciple: :shrug: i implemented based on the docs for [GitHub rate limit rules](https://developer.github.com/v3/#rate-limiting) (403). But you are right, based on some light searching 429 may be more common rate limit error code. I can update the code to handle both. At most will improve the error message; will still fallback either way.

src/installer.ts
bryanmacfarlane: The toolkit is merged and released.  Do you want to update this to pass `token: ${token}`?

LocalCragDirtBag (commented):
gh pr checkout 157

Issue #160, opened 2020-06-04
Reduce latency when new node versions are released


kenhuang:
Hi there:

Thanks for this library. Just found that the ` node-version` option is not downloading the latest node release.

v10.21.0 release notes:
https://nodejs.org/en/blog/vulnerability/june-2020-security-releases/

Our setup:
```
- uses: actions/setup-node@v1
  with:
    node-version: 10.x
    registry-url: "https://registry.npmjs.org"
```

Logs:
```
Run actions/setup-node@v1
/usr/bin/docker exec  be3164773b20b9e565b79dfa087a0d06395e9681e0de57453920d34a25c07234 sh -c "cat /etc/*release | grep ^ID"
/__t/node/10.20.1/x64/bin/node --version
v10.20.1
/__t/node/10.20.1/x64/bin/npm --version
6.14.4
```

Update:
seems related to this change: https://github.com/actions/setup-node/pull/147

Regards
Ken

gabrielschulhof:
The same thing is true when specifying "v12.x". If one wants the latest version one must specify it explicitly. I needed something later than the version 12.16.3 it was grabbing when given "v12.x", so I wrote "v12.18.0" for my node-version. This is not that great though, since I wish to always use the latest release.

gabrielschulhof:
https://github.com/nodejs/node-addon-examples/pull/139#issuecomment-640749751

maxim-lobanov:
Hello everyone, just to clarify how it works:
1) `actions/setup-node` is trying to find version that satisfy your version spec on image
2) If it is found - it is applied (to avoid spending time for downloading).
3) If it is not found, it downloads latest satisfied version and install it in runtime.

Unfortunately, pre-cached versions are updated with small delay (usually, not longer than 1-1.5 weeks because images rollout takes much effort) so if you specify full version, action doesn't find it on image and downloads and installs it in runtime. If you just specify `10.x`, it resolves to pre-cached version that can be outdated due to image.

@bryanmacfarlane , may be some flag similar to `checkLatest` in Azure DevOps UseNode task can be implemented to fix this problem. If `checkLatest` flag is `true`, we determine latest version firstly, search it on image and download if it is not found. It will help in cases where 1-2 weeks delay is not acceptable.

bryanmacfarlane:
[...]

PR #165, opened 2020-06-22
Implement "check-latest" flag to check if pre-cached version is latest one

maxim-lobanov:
Issue: https://github.com/actions/setup-node/issues/160

Add `check-latest` flag and implementation

Additional changes:
1) Add tests
2) Significantly extend test matrix in `versions.yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,19 @@ steps:
 - run: npm test
 ```
 
+Checking latest version of Node.js:  
+(By default, action searches version locally before downloading it. This flag forces action to check if local version is the latest one)
+```yaml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-node@v2
+  with:
+    node-version: '12'
+    check-latest: true
+- run: npm install
+- run: npm test
+```
+
 Matrix Testing:
 ```yaml
 jobs:
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ steps:
 ```
 
 Checking latest version of Node.js:  
-(By default, action searches version locally before downloading it. This flag forces action to check if local version is the latest one)
+(By default, action searches version locally before downloading it. The flag forces action to check if cached version is not outdated)
 ```yaml
 steps:
 - uses: actions/checkout@v2
--- a/README.md
+++ b/README.md
@@ -41,8 +41,9 @@ steps:
 - run: npm test
 ```
 
-Checking latest version of Node.js:  
-(By default, action searches version locally before downloading it. The flag forces action to check if cached version is not outdated)
+Check latest version:  
+> In basic example, without `check-latest` flag, the action tries to resolve version from local cache firstly and download only if it is not found. Local cache on image is updated with a couple of weeks latency.  
+`check-latest` flag forces the action to check if the cached version is the latest one. It reduces latency significantly but it is much more likely to incur version downloading.
 ```yaml
 steps:
 - uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


konradpabjan:
I think some updates to the `README` regarding the new input would be good. Not everyone looks at `action.yml` to get the list of all available inputs.

An example with `check-latest` set to true would be good alongside an explanation of why users would want to us this

src/installer.ts
```diff
@@ -141,6 +114,51 @@ export async function getNode(
   core.addPath(toolPath);
 }
 
+async function installNodeVersion(
+  info: INodeVersionInfo,
+  auth: string | undefined
+): Promise<string> {
+  const localPath = tc.find('node', info.resolvedVersion);
+  if (localPath) {
+    console.log(`Found in cache @ ${localPath}`);
```
konradpabjan, line 123: Instead of `console.log`, you can use `core.info` for tracing. It's effectively the same thing but it's just fits better with the rest of the code. 

For some reason it isn't documented, a PR is on the way to update the README
https://github.com/actions/toolkit/blob/9e2d61e548a6edd163b05e31b69a8639417cdcac/packages/core/src/core.ts#L153

maxim-lobanov: Just copy-pasted existing logging. Not sure why it was done initially. I have replaced console.log to core.info through all code

maxim-lobanov:
> Some updates to the `README` regarding the new input would be good. Not everyone looks at `action.yml` to get the list of all available inputs.
> 
> An example with `check-latest` set to true would be good alongside an explanation of why users would want to us this (what specific use cases does this solve or fix). Here you can also mention the default behavior.

@konradpabjan , Updated readme. Please let me know if we can find better words to explain flag purpose

src/installer.ts
```diff
@@ -26,20 +26,36 @@ interface INodeVersionInfo {
...
 
+  if (checkLatest) {
+    core.info('Attempt to resolve the latest version from manifest...');
+    const resolvedVersion = await resolveVersionFromManifest(
+      versionSpec,
+      stable,
+      auth
+    );
+    if (resolvedVersion) {
+      versionSpec = resolvedVersion;
```
bryanmacfarlane, line 43: cool.  so just to make 100% sure, if checkLatest, we resolve the inputted version spec against the node-versions cache manifest, and then rewrite the versionSpec as the resolved version.  I like that.

src/installer.ts
```diff
@@ -197,6 +210,20 @@ async function getInfoFromDist(
...
+async function resolveVersionFromManifest(
+  versionSpec: string,
+  stable: boolean,
+  auth: string | undefined
+): Promise<string | undefined> {
+  try {
+    const info = await getInfoFromManifest(versionSpec, stable, auth);
+    return info?.resolvedVersion;
+  } catch (err) {
+    core.warning('Unable to resolve version from manifest...');
```
bryanmacfarlane, line 222: Should this be a warning?  A warning means the user should take action.  I think we should just output (core.info) so the log is clear what happened but no need for worry and warning.  Seems that's appropriate for later downstream when ultimately a version can't be resolved and that's currently handled.

README.md
```diff
@@ -41,6 +41,19 @@ steps:
 - run: npm test
 ```
 
+Checking latest version of Node.js:
```
bryanmacfarlane, line 44: I think we should add a note about it being a tradeoff between perf / reliability and ensuring latest.  Default of check-latest false ensure instant resolution to major / minor version with up to a couple of weeks latency.  Check-latest true reduces latency significantly but is much more likely to incur a download and extract.   Saying that will help users make the trade-off.  I want to make sure they think about how badly they want that reduced latency.

bryanmacfarlane:
I added a few notes.  Outside of that, I'm good with the change.

README.md
maxim-lobanov: I have added a few more sentences. Please let me know if wording should be changed.

PR #171, opened 2020-07-15
Switch from "master" to "main" branch

maxim-lobanov:
Recently, we have switched default branch in https://github.com/actions/node-versions to `main` and we need to switch this action too.

This fix should be integrated to tags `v2-beta`, `v2.0.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ steps:
     node-version: '12'
 ```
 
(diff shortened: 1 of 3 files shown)


bryanmacfarlane:
Looks good, but I would add a note in this section where we refer to the node-versions repo and querying versions.  Make sure it's clear that we will query main.  I will take out of beta after this.  https://github.com/actions/setup-node#usage

PR #178, opened 2020-08-06
Update Third Party License Information

thboop:


bryanmacfarlane:
I believe main is v2-beta and not going into ghes ?? there is a v1 branch.  Should this PR target that?

PR #184, opened 2020-08-15
chore: operating-system -> os

nschonni:
Schema validation was flagging the old property name

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -13,10 +13,10 @@ on:
       
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


bryanmacfarlane:
which schema validation?  os isn't a well known property - the matrix leg name can be whatever we want.  was it because of the hyphen?

nschonni:
Through VS Code, but I don't really feel strongly about this, since either seems to work. I think `os` is just more commonly used in examples and snippets.

bryanmacfarlane:
cool.  merged.  just wanted to understand

PR #185, opened 2020-08-15
refactor: use core.info instead of console.log

rchougule:
Use `core.info` instead of `console.log` for keeping it consistent with other files.


bryanmacfarlane:
Thanks

PR #186, opened 2020-08-15
chore: Remove strategy for non-matrix builds

nschonni:
Just flagged in schema validation because the strategy value is only applicable if it has a matrix below it

### DIFF
--- a/.github/workflows/proxy.yml
+++ b/.github/workflows/proxy.yml
@@ -3,14 +3,14 @@ name: proxy
 on:
   pull_request:
     paths-ignore:
-      - '**.md'    
-  push:    
+      - '**.md'
+  push:
     branches:
-      - master
+      - main
       - releases/*
     paths-ignore:
       - '**.md'
-      
+
 jobs:
   test-proxy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bryanmacfarlane:
Is the schema validation wrong?  Even though there's no matrix below there are multiple jobs and the intent is to fail fast the other jobs if one fails to release concurrency.  Not sure this is a good change but maybe I'm missing something ...

nschonni:
The separate `job`s aren't part of a matrix. EX: if one fails it doesn't fail/cancel the others. The matrix is used to pass variables to them. If the one of the "batches" like running Node 14 and Node 12 fails, then it cancels all the jobs in the other batch, not other jobs in the same batch.

nschonni:
Found the schema that VS Code is using https://json.schemastore.org/github-workflow

bryanmacfarlane:
It might be a bug in the schema - in this case we want fail fast even though it's not a matrix.

nschonni:
Did an example to show that it doesn't behave like you're talking about https://github.com/nschonni/setup-node/actions/runs/499393700
Even if one of the job fails, it doesn't cancel other running jobs like it does when they are part of an actual matrix

theifedayo (commented):
what does strategy: failfast: false do? can you explain?

MaksimZhukov (approved):

dmitry-shibanov (approved):

PR #200, opened 2020-09-29
Swap to env files

thboop:
The newest toolkit version uses [Environment files](https://github.com/actions/toolkit/blob/main/docs/commands.md#environment-files) to communicate with the Runner. This pr moves towards the new way of communicating with the runner and updates the toolkit version if appropriate.


bryanmacfarlane:
Probably good to bump the minimum here to ^1.2.6 https://github.com/actions/setup-node/blob/main/package.json#L26

bryanmacfarlane:
Outside of that, looks good.

bryanmacfarlane (changes requested):
feel free to merge after the suggested change

Issue #32, opened 2019-08-15
Support .nvmrc


dougmoscrop:
It would be nice if `setup-node` understood an .nvmrc file (even if it does not use nvm under the hood, the idea is the `node-version` is just the contents of the file).

If GitHub Actions as a whole could read the contents of a file as part of an expression, I could do something like

```yml
uses: setup-node@v1
with:
  node-version: {{ contents(.nvmrc) }}
```

But I don't know where to log that type of feature request!

Perhaps '.nvmrc' could be used as a special placeholder, so that:

```yml
uses: setup-node@v1
with:
  node-version: '.nvmrc'
```

Instructs this action to read the contents and treat the contents as the desired version?

damccorm:
> If GitHub Actions as a whole could read the contents of a file as part of an expression

I think its probably unlikely we're going to expand our expression parsing to that degree - in general computation isn't supposed to happen there and we want to discourage moving in that direction, and keep that more towards just reading existing variables. Note that you could do something like the following here:

```
run: nvmrc=$(cat .nvmrc)

uses: setup-node@v1
with:
  node-version: $nvmrc
```

> Perhaps '.nvmrc' could be used as a special placeholder

I'm interested in this option. We're looking at doing something similar in setup-dotnet: https://github.com/actions/setup-dotnet/pull/15. I see a couple options here:

1) If there's a `.nvmrc` at the root of the repo (or the current working directory) and no `node-version` specified, automatically use the version in the `.nvmrc` file. Maybe have a variable that allows you to turn this on or off.
2) Allow `node-version` to be a path to a `.nvmrc` file.
3)  Don't allow this option.

@chrispat, any thoughts on this one? I think maybe we should establish a pattern here and with setup-dotnet for similar requests (I'm sure node and dotnet aren't the only tools with config files that can specify versions).

dougmoscrop:
Thanks for the top about just using an output, that makes sense and is good enough for me, though the simpler developer experience would be nice;  maybe it's just worth a `setup-npm` action?

travi:
> I see a couple options here:
>
> 1. If there's a .nvmrc at the root of the repo (or the current working directory) and no node-version specified, automatically use the version in the .nvmrc file. Maybe have a variable that allows you to turn this on or off.
> 1. Allow node-version to be a path to a .nvmrc file.
> 1. Don't allow this option.

[...]

PR #209, opened 2020-11-01
Node version file support

TJMcCarthy95:
### Overview

Implemented the ability to define a file within a project as the holder of the projects node version such as an `.nvmrc` file so to mitigate duplicating node version declarations. After recently researching the ability to do this I came across this ["issue"](https://github.com/actions/setup-node/issues/32) (albeit not an issue) and then came across this [comment](https://github.com/actions/setup-node/issues/32#issuecomment-522083355) which is the implementation I had envisaged.

The implementation follows a similar flow as defined in the comment whereby the `node-version` input takes precedence and if not provided and the `node-version-file` input is, then read the provided node version file.

### Commits

- Implemented support for repository defined node version files such as '.nvmrc'

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,21 @@ steps:
 - run: npm test
 ```
 
+Node version file:
+
+The `node-version-file` input allows you to use a file within your repository which contains the version of node your project uses for example `.nvmrc`. If both the `node-version` and the `node-version-file` inputs are provided the `node-version` input is used.
+> The node version file is read from the project root
+
+```yaml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-node@v2
+  with:
+    node-version-file: '.nvmrc'
+- run: npm install
+- run: npm test
+```
+
 Matrix Testing:
 ```yaml
 jobs:
--- a/__tests__/installer.test.ts
+++ b/__tests__/installer.test.ts
@@ -36,6 +36,7 @@ describe('setup-node', () => {
   let dbgSpy: jest.SpyInstance;
   let whichSpy: jest.SpyInstance;
   let existsSpy: jest.SpyInstance;
+  let readFileSyncSpy: jest.SpyInstance;
   let mkdirpSpy: jest.SpyInstance;
   let execSpy: jest.SpyInstance;
   let authSpy: jest.SpyInstance;
@@ -67,6 +68,7 @@ describe('setup-node', () => {
     // io
     whichSpy = jest.spyOn(io, 'which');
     existsSpy = jest.spyOn(fs, 'existsSync');
+    readFileSyncSpy = jest.spyOn(fs, 'readFileSync');
     mkdirpSpy = jest.spyOn(io, 'mkdirP');
 
     // disable authentication portion for installer tests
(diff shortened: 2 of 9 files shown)


sangeeth96:
Hey @TJMcCarthy95, thanks for this PR! I intended to take up this PR but got sidetracked with work. Few things I wanted to point out.

First one is that `.nvmrc` accepts formats like the following:

- partial versions such as `12` or `12.16` (I believe this is handled right now?)
- `lts/*` for the latest LTS release
- `lts/erbium` for specific LTS release

and probably more. You can find this via `nvm --help`. I don't think this PR addresses those as well, I wrote some tests locally which failed.

There's also the problem of `.nvmrc` being just a feature of `nvm` and not being a standard. There was an attempt to standardize on a format for all of Node.js in nodejs/version-management#13 but that didn't take off as expected. There's also the `.node-version` file syntax, which is much simpler, and supported by multiple tools listed in https://github.com/shadowspawn/node-version-usage. So, I'd also suggest changing the input variable name to `nvmrc-file` to probably avoid confusion (or maybe not have that). Or, the tool could support reading such files as well (should be more straightforward than the nuances that come with `.nvmrc`).

I don't suppose we need to take into account every `.nvmrc` syntax since the purpose here is just to fetch a compatible Node.js installation. However, I'd say that it'd be useful to support the above mentioned `.nvmrc` formats. For eg, some of the repos I work on use `lts/erbium` to make sure the corresponding latest LTS release is used every time CI is run.

If you're interested, I do have a branch based off of this PR with the changes required to parse the above syntaxes.

TJMcCarthy95:
> Hey @TJMcCarthy95, thanks for this PR! I intended to take up this PR but got sidetracked with work. Few things I wanted to point out.
> 
> First one is that `.nvmrc` accepts formats like the following:
> 
> * partial versions such as `12` or `12.16` (I believe this is handled right now?)
> * `lts/*` for the latest LTS release
> * `lts/erbium` for specific LTS release
> 
> and probably more. You can find this via `nvm --help`. I don't think this PR addresses those as well, I wrote some tests locally which failed.
> 
> There's also the problem of `.nvmrc` being just a feature of `nvm` and not being a standard. There was an attempt to standardize on a format for all of Node.js in [nodejs/version-management#13](https://github.com/nodejs/version-management/issues/13) but that didn't take off as expected. There's also the `.node-version` file syntax, which is much simpler, and supported by multiple tools listed in https://github.com/shadowspawn/node-version-usage. So, I'd also suggest changing the input variable name to `nvmrc-file` to probably avoid confusion (or maybe not have that). Or, the tool could support reading such files as well (should be more straightforward than the nuances that come with `.nvmrc`).
> 
> I don't suppose we need to take into account every `.nvmrc` syntax since the purpose here is just to fetch a compatible Node.js installation. However, I'd say that it'd be useful to support the above mentioned `.nvmrc` formats. For eg, some of the repos I work on use `lts/erbium` to make sure the corresponding latest LTS release is used every time CI is run.
> 
> If you're interested, I do have a branch based off of this PR with the changes required to parse the above syntaxes. You can check it out [here](https://github.com/sangeeth96/setup-node/tree/add-node-version-file-support--additional-syntaxes).

I appreciate you taking the time to review the PR. My intentions of implementing this weren't necessarily to support the complete feature set of `.nvmrc` but rather just provide the ability to have a node version defined in a file, the `.nvmrc` just happens to be my usage, which is the primary reason I didn't tie any variable names to nvm. I could make the `README` to be more explicit about that, or rather change the example to be more generic although I like what you've put together so might update accordingly.

TJMcCarthy95:
@sangeeth96 Sorry for the delay on these updates got caught up with work. I've implemented the updates using your branch as reference so when you get a chance, would be great if you could give it another review, cheers.

jasonkarns:
Just commenting since it's highly related: https://github.com/nodenv/actions-node-version is an action that reads .node-version file (leveraging [nodenv](https://github.com/nodenv/nodenv)) and exposes it as an output that can be used as an input to the setup-node action. This would be a workaround until/unless setup-node gains the ability to do it automatically.

Cu2y (approved):

hkaur008:
Is this branch deleted @alepauly  ?

PR #222, opened 2021-01-01
Read version from .nvmrc or .tool-versions config

matoakley:
[NVM](https://github.com/nvm-sh/nvm) and [asdf](https://github.com/asdf-vm/asdf) are popular tools for managing node (and other dependency) versions in projects. These tools provide the ability to specify the version number in either a `.nvmrc` or `.tool-versions` file and have the management tool automatically install and configure that version.

This patch (shamelessly plagiarised and adapted from [setup-ruby](https://github.com/ruby/setup-ruby)) will attempt to set the version from the `.nvmrc` or `.tool-versions` file if not specified explicitly. If it is unable to find and parse a version then it will return a blank string and fallback to the original logic to use the architecture's pre-installed node.

To be clear, the order of precedence is:

1. Use the version explicitly set in the `version` or `node-version` option,
2. Attempt to read the version from `.nvmrc` if the file is present,
3. Attempt to read the version from `.tool-versions` if the file is present,
4. Fallback to the system node.

By enabling the action to read from these common configuration files we reduce coupling and allow developers to manage their node version from a single location within their codebase.

This patch provides the enhancement discussed in #32 .

Finally, this is the first time that I've worked with Typescript (I'm a rubyist at heart) so I'm open to code review and suggestions to improve the code itself.

Happy New Year! 🎉

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -4695,6 +4695,7 @@ const auth = __importStar(__webpack_require__(202));
 const path = __importStar(__webpack_require__(622));
 const url_1 = __webpack_require__(835);
 const os = __webpack_require__(87);
+const fs = __webpack_require__(747);
 function run() {
     return __awaiter(this, void 0, void 0, function* () {
         try {
@@ -4702,10 +4703,7 @@ function run() {
             // Version is optional.  If supplied, install / use from the tool cache
             // If not supplied then task is still used to setup proxy, auth, etc...
             //
-            let version = core.getInput('node-version');
-            if (!version) {
-                version = core.getInput('version');
-            }
+            let version = parseNodeVersion();
             let arch = core.getInput('architecture');
             // if architecture supplied but node-version is not
             // if we don't throw a warning, the already installed x64 node will be used which is not probably what user meant.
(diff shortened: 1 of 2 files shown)


dist/index.js
```diff
@@ -4742,6 +4740,29 @@ function isGhes() {
     const ghUrl = new url_1.URL(process.env['GITHUB_SERVER_URL'] || 'https://github.com');
     return ghUrl.hostname.toUpperCase() !== 'GITHUB.COM';
 }
+function parseNodeVersion() {
+    let nodeVersion = core.getInput('node-version') || core.getInput('version');
```
jamiehaywood, line 4744: just wanted to ask here whether you want to check for a `falsy` value or just `null` or `undefined`? If the latter:

```suggestion
    let nodeVersion = core.getInput('node-version') ?? core.getInput('version');
```

src/main.ts
```diff
@@ -64,3 +61,29 @@ function isGhes(): boolean {
...
+      // Read from .nvmrc
+      nodeVersion = fs.readFileSync('.nvmrc', 'utf8').trim();
+      console.log(`Using ${nodeVersion} as input from file .nvmrc`);
+    } else if (fs.existsSync('.tool-versions')) {
+      // Read from .tool-versions
+      const toolVersions = fs.readFileSync('.tool-versions', 'utf8').trim();
+      const nodeLine = toolVersions
+        .split(/\r?\n/)
+        .filter(e => e.match(/^nodejs\s/))[0];
+      nodeVersion = nodeLine.match(/^nodejs\s+(.+)$/)![1];
```
jamiehaywood, line 79: `.match()` returns `RegExpMatchArray | null`, so rather than using definite assignment assertion (`!`), perhaps use the optional chain (`?`) to query the RegExpMatchArray. If nothing is returned here, it will assign `undefined` to `nodeVersion`. 

This does then throw into question the flow of these statements, because if I have a `.tool-versions` with the `nodejs` key in (I think that's what you're testing?) but no value, then it will return undefined.

I would revisit this by allowing TS to guide you with the types & then making sure that a string is definitely being returned.

```suggestion
      nodeVersion = nodeLine.match(/^nodejs\s+(.+)$/)?.[1];
```

jasonkarns:
Looks like a dupe of #209 ?

Just commenting since it's highly related: https://github.com/nodenv/actions-node-version is an action that reads .node-version file (leveraging [nodenv](https://github.com/nodenv/nodenv)) and exposes it as an output that can be used as an input to the setup-node action. This would be a workaround until/unless setup-node gains the ability to do it automatically.

ruscon:
@matoakley @jamiehaywood Is there anything I can do to complete this feature?

jamiehaywood:
@ruscon unfortunately I'm not a maintainer, I just added a review because I'm keen to get this feature into `setup-node`. I think there are 2 or 3 PRs covering the same feature floating around, so it might be worth having a look on them as well, and seeing the status of that.

ruscon:
this solution is better
https://github.com/actions/setup-node/pull/239

matoakley:
@jamiehaywood Thanks for the review it's much appreciated 👍 

Sadly I've lost hope on this PR given that the project appears to be all but abandoned by its maintainers. As has been pointed out, there are now at least 3 open PRs offering a solution to this problem and there seems to be high demand for the feature.

@ruscon I agree that #239 would probably be a better all-round solution but whether it is "better" is subjective. For example, it wouldn't have solved my problem as the patch does not lookup the `.tool-versions` file which https://github.com/asdf-vm/asdf uses.

In the 3 months since I submitted this PR I have migrated my development environment to Docker and have resigned myself to managing the Node version manually for Github Actions.

PR #226, opened 2021-01-15
Update action.yml

hubsen1980:


bryanmacfarlane:
no context on the change and why it's needed.  closing for now.

PR #233, opened 2021-02-05
better understand why set env NODE_AUTH_TOKEN . and what used for

tingyu22:
didn't see anywhere explain that,  make us duplicate use another token finish publishing work

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ jobs:
       - run: npm test
 ```
 
-Publish to npmjs and GPR with npm:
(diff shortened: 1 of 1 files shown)


dmitry-shibanov:
Hello @tingyu22. Thank you for your pull request. I think it's not necessary, because `NODE_AUTH_TOKEN` environment variable has clear name. Besides, for more description about environment variable generated in scope of `registry-url` input you can see [here](https://github.com/actions/setup-node/blob/main/action.yml) and GitHub [docs](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry).

PR #247, opened 2021-03-29
Add npm-version option

frbuceta:
This change adds an option which allows to install a different version of NPM


lastthyme (approved):

mattstrayer:
This would be a very welcomed feature. Can we get this out the door?

mellis481:
@frbuceta Are you able to see this PR through please?  Would be great to get this in.

pierresisson (approved):
LGTM

sztadii:
@frbuceta would you like to resolve the conflicts? I was thinking to implement this feature by myself, but in this case, you will lose the whole glory :(

sztadii:
It looks like the forked repo does not exist, so we can not resolve the conflicts. Should we close the PR?

brcrista:
@sztadii yes, we'll have to do that to make progress.

Looks like a useful feature.  But rather than adding a separate input, perhaps we should just support [`engines.npm`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines) as we are with `engines.node`: https://github.com/actions/setup-node/pull/485

PR #266, opened 2021-05-21
Create ADR for integrating cache functionality to setup-node action

maxim-lobanov:


docs/adrs/0000-caching-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+# Context
+`actions/setup-node` is the 2nd most popular action in GitHub Actions. A lot of customers use it in conjunction with [actions/cache](https://github.com/actions/cache) to speed up dependencies installation.  
+See more examples on proper usage in [actions/cache documentation](https://github.com/actions/cache/blob/main/examples.md#node---npm).
+
+# Goals & Anti-Goals
+Integration of caching functionality into `actions/setup-node` action will bring the following benefits for action users:
+- Decrease the entry threshold for using the cache for Node.js dependencies and simplify initial configuration
+- Simplify YAML pipelines because no need additional steps to enable caching
+
+As a result, more users will use the cache for Node.js builds and will be happy with fast builds.
```
AlenaSviridenko, line 15: I would move speed up to the benefits list, because it can save cost of the workflow

docs/adrs/0000-caching-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+
+We don't persue the goal to provide wide customization of caching in scope of `actions/setup-node` action. The purpose of this integration is covering ~90% of basic use-cases. If user needs flexible customization, we should advice them to use `actions/cache` directly.
+
+# Decision
+- Add `cache` input parameter to `actions/setup-node`. For now, input will accept the following values: 
+  - `npm` - enable caching for npm dependencies
+  - `''` - disable caching (default value)
+  - Potentially, we will be able to extend this input to support Yarn
+- Cache feature will be disabled by default to make sure that we don't break existing customers. We will consider enabling cache by default in next major release (`v3`)
+- Add optional input `package-lock-path` that will allow to specify path to `package.lock.json` file path:
```
AlenaSviridenko, line 26: package-lock.json is the correct naming. It should be updated in several places

AlenaSviridenko (approved):

MaksimZhukov (approved):

docs/adrs/0000-caching-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+- Decrease the entry threshold for using the cache for Node.js dependencies and simplify initial configuration
+- Simplify YAML pipelines because no need additional steps to enable caching
+- More users will use cache for Node.js so more customers will have fast builds!
+
+As the first stage, we will add support for NPM dependencies caching. We can consider adding the same functionality for Yarn later.
+
+We don't persue the goal to provide wide customization of caching in scope of `actions/setup-node` action. The purpose of this integration is covering ~90% of basic use-cases. If user needs flexible customization, we should advice them to use `actions/cache` directly.
+
+# Decision
+- Add `cache` input parameter to `actions/setup-node`. For now, input will accept the following values:
```
maxim-lobanov, line 21: We have thought about making this input boolean like `cache: true` and it would work for us for now. But if we want to extend caching to Yarn in future, we will need to add additional field to choose between Yarn / npm. Npm 7 started t support `yarn.lock` so we can't choose package manager by filename anymore.

Any thoughts and feedback are appreciated

docs/adrs/0000-caching-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+As the first stage, we will add support for NPM dependencies caching. We can consider adding the same functionality for Yarn later.
+
+We don't persue the goal to provide wide customization of caching in scope of `actions/setup-node` action. The purpose of this integration is covering ~90% of basic use-cases. If user needs flexible customization, we should advice them to use `actions/cache` directly.
+
+# Decision
+- Add `cache` input parameter to `actions/setup-node`. For now, input will accept the following values: 
+  - `npm` - enable caching for npm dependencies
+  - `''` - disable caching (default value)
+  - Potentially, we will be able to extend this input to support Yarn
+- Cache feature will be disabled by default to make sure that we don't break existing customers. We will consider enabling cache by default in next major release (`v3`)
```
maxim-lobanov, line 25: Considering the fact that users can use either `npm` or `yarn`, I am not sure actually how we can enable it by default in future without making `cache` input mandatory (that doesn't sound good from UX side).
@ahdbilal , any concerns about it?

docs/adrs/0000-caching-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+- Add optional input `package-lock-path` that will allow to specify path to `package-lock.json` file path:
```
maxim-lobanov, line 26: @ahdbilal , I am not sure what is the best name for this input instead of `package-lock-path`.
The purpose of the input is specifying the path to file which hash will be used as caching key

docs/adrs/0000-caching-dependencies.md
ahdbilal: fine with me for not making this on by default. I will get other folks opinions on it.

docs/adrs/0000-caching-dependencies.md
ahdbilal: maybe `working-directory`. I will get others' opinions on it.

maxim-lobanov: This input should accept either directory path or file path so `working-directory` doesn't sound correct. Also, `working-directory` sounds very generic. It is not clear what should be put in this input

docs/adrs/0000-caching-dependencies.md
ahdbilal: I really want this to be boolean. can we not automatically detect the package manager?

docs/adrs/0000-caching-dependencies.md
ahdbilal: agreed! let us get others' opinions on it.

docs/adrs/0000-caching-dependencies.md
maxim-lobanov: No way unfortunately, because we can't rely on file name.
I agree on having `cache: boolean` but in this case, we will need additional field in future like `package-manager`

konradpabjan: I think `cache: npm` is the best name that we can then re-use across some of our other first party actions.

For example long-term with `setup-python`, there are a bunch of package managers like `pip`, `conda` and `virtualenv` that would have to be supported so having `cache: true/false` would be very hard to extend

docs/adrs/0000-caching-dependencies.md
konradpabjan: This statement seems appropriate 👍 No need to change the current behavior with `v2`.

We can initially try it out and if it works really well and users see a large improvement then we can consider trying to enable this for `v3`.

One option with `v3` is to make the `cache` input mandatory but we set the default value to `npm`. If there is no `package-lock.json` then the action will output a warning instead of failing. This will not break existing workflows, the majority of users would see improvements due to utilizing cache and for `yarn` users they would only see a warning annotation that should be descriptive enough for them to figure out what is going on. With a mandatory input though, we could add something like a `cache: disabled` option for users who really don't want it. In the future we can make that call but for the moment making it optional is perfect just so it can be tested out.

One potential issue I could see happening with caching being mandatory when moving to v3 is that some users might have extra caching happening that causes them to use more than 5GB. That could cause automatic eviction and all sorts of headaches so we would need to document things extensively before. Lots of considerations but in general I think we should be aggressive about trying to enabling caching by default.

docs/adrs/0000-caching-dependencies.md
konradpabjan: `cache-package-path` might be worth considering. For all the big setup actions, python, dotnet, node I think we would be primarily be focusing on speeding up caching of dependencies that in some way originate from **package managers**. For node that is `npm` and `yarn`, for Python this is `conda`, `pip`, for Dotnet this could be `nuget` or potentially `Chocolatey NuGet`.

I like having `cache` at the start because then it's immediately obvious that this is in some way related to caching and the earlier `cache` input. If we remove the `-lock` from this input then it becomes generic enough where we can reuse this across all of our first party actions.

`cache-dependency-path` is another option, but I don't like it that much. My vote goes to `cache-package-path`

docs/adrs/0000-caching-dependencies.md
maxim-lobanov: That is really good point

docs/adrs/0000-caching-dependencies.md
maxim-lobanov: I would rather keeping cache disabled even in v3  because don't want to push customers choosing anything. In setup-java, we have seen a lot of feedback that customers don't want to choose between adopt and zulu

alepauly (approved):
The general idea makes sense to me. Left a couple of comments

docs/adrs/0000-caching-dependencies.md
maxim-lobanov: In my opinion, the providing ability to specify package lock path should be available in the first iteration. It is pretty difficult from implementation side and don't have underwater rocks.
Without this feature, we will drop all use-cases where package-lock.json / yarn.lock are located in any subfolder (that looks like popular case).

It is very very basic customization that "should be". Just my IMHO.

As for the naming of field, I would prefer `package-lock-path` because it is clearer what should be specified. But i am open for suggestions

maxim-lobanov: We have decided to not support custom location for `package-lock.json` and `yarn.lock` in the first iteration

AlenaSviridenko (approved):

dmitry-shibanov (approved):

PR #270, opened 2021-06-17
Support LTS aliases

gordey4doronin:
The work is related to both https://github.com/actions/setup-node/issues/26 and https://github.com/actions/setup-node/issues/32.

For manifest changes see https://github.com/actions/versions-package-tools/pull/32 and https://github.com/actions/node-versions/pull/63.

- [x] Support `lts/codename` aliases
- [x] Support `lts/*` alias
- [x] Test coverage for `lts/codename` aliases
- [x] Test coverage for `lts/*` alias
- [x] Test coverage for malformed aliases
- [x] Clarify whether I need to commit compiled dist files or not? CC @maxim-lobanov 
- [x] Discuss whether forcing `check-latest = true` is good idea or not? CC @maxim-lobanov 
- [x] Discuss whether it's worth adding support for simple `erbium` or `lts/erbium` is enough? CC @maxim-lobanov @konradpabjan @bryanmacfarlane 

The goal of my PR is to support `.nvmrc` syntax.
**Given**: `.nvmrc`/`.node-version` file in a repository
**When**: run `setup-node` action and pass content of the file to `version` property
**Then**: the action can recognize the format and choose correct node version

Being said that, from my point of view supporting simple `erbium` is not required, since version manager doesn't allow that.
Also, having all codenames starting with `lts/` makes it really easy to determine in the code when `lts` alias passed to the action.

---

Reading the `.nvmrc` file is not the goal of this PR.
A separate PR shall be created for that, but it wouldn't be possible to implement `.nvmrc` reading without supporting the format.
So, think about current PR as a pre-requisite. 🙂

### DIFF
--- a/__tests__/installer.test.ts
+++ b/__tests__/installer.test.ts
@@ -399,6 +399,7 @@ describe('setup-node', () => {
       expect(logSpy).not.toHaveBeenCalledWith(
         'Attempt to resolve the latest version from manifest...'
       );
+      expect(dbgSpy).not.toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
     });
 
     it('check latest version and resolve it from local cache', async () => {
@@ -419,6 +420,7 @@ describe('setup-node', () => {
       expect(logSpy).toHaveBeenCalledWith(
         'Attempt to resolve the latest version from manifest...'
       );
+      expect(dbgSpy).toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
       expect(logSpy).toHaveBeenCalledWith("Resolved as '12.16.2'");
       expect(logSpy).toHaveBeenCalledWith(`Found in cache @ ${toolPath}`);
     });
@@ -443,6 +445,7 @@ describe('setup-node', () => {
       expect(logSpy).toHaveBeenCalledWith(
         'Attempt to resolve the latest version from manifest...'
       );
+      expect(dbgSpy).toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
       expect(logSpy).toHaveBeenCalledWith("Resolved as '12.16.2'");
       expect(logSpy).toHaveBeenCalledWith(
         `Acquiring 12.16.2 - ${os.arch} from ${expectedUrl}`
@@ -479,6 +482,7 @@ describe('setup-node', () => {
       expect(logSpy).toHaveBeenCalledWith(
         'Attempt to resolve the latest version from manifest...'
       );
+      expect(dbgSpy).toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
       expect(logSpy).toHaveBeenCalledWith(
         `Failed to resolve version ${versionSpec} from manifest`
       );
@@ -522,6 +526,7 @@ describe('setup-node', () => {
       );
       expect(logSpy).toHaveBeenCalledWith(
         'Unable to resolve version from manifest...'
+        // 'Unable to get manifest...'
       );
       expect(logSpy).toHaveBeenCalledWith(
         `Failed to resolve version ${versionSpec} from manifest`
@@ -549,11 +554,11 @@ describe('setup-node', () => {
       await main.run();
 
       // assert
-      expect(logSpy).toHaveBeenCalledWith('LTS version is provided. For LTS versions `check-latest` will be automatically set to true')
-      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve the latest version from manifest...');
+      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve LTS alias from manifest...')
+      expect(dbgSpy).toHaveBeenCalledWith('Getting manifest from actions/node-versions@main')
+      expect(dbgSpy).not.toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
       expect(dbgSpy).toHaveBeenCalledWith(`LTS alias 'erbium' for Node version 'lts/erbium'`)
       expect(dbgSpy).toHaveBeenCalledWith(`Found LTS release '12.16.2' for Node version 'lts/erbium'`)
-      expect(logSpy).toHaveBeenCalledWith("Resolved as '12.16.2'");
       expect(logSpy).toHaveBeenCalledWith(`Found in cache @ ${toolPath}`);
       expect(cnSpy).toHaveBeenCalledWith(`::add-path::${toolPath}/bin${osm.EOL}`);
     });
@@ -577,12 +582,12 @@ describe('setup-node', () => {
       await main.run();
 
       // assert
-      expect(logSpy).toHaveBeenCalledWith('LTS version is provided. For LTS versions `check-latest` will be automatically set to true')
-      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve the latest version from manifest...');
+      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve LTS alias from manifest...')
+      expect(dbgSpy).toHaveBeenCalledWith('Getting manifest from actions/node-versions@main')
+      expect(dbgSpy).not.toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
       expect(dbgSpy).toHaveBeenCalledWith(`LTS alias 'erbium' for Node version 'lts/erbium'`)
       expect(dbgSpy).toHaveBeenCalledWith(`Found LTS release '12.16.2' for Node version 'lts/erbium'`)
-      expect(logSpy).toHaveBeenCalledWith("Resolved as '12.16.2'");
-      expect(logSpy).toHaveBeenCalledWith("Attempting to download 12.16.2...");
+      expect(logSpy).toHaveBeenCalledWith("Attempting to download 12...");
       expect(logSpy).toHaveBeenCalledWith(`Acquiring 12.16.2 - ${os.arch} from ${expectedUrl}`);
       expect(logSpy).toHaveBeenCalledWith('Extracting ...');
       expect(logSpy).toHaveBeenCalledWith('Adding to the cache ...');
@@ -604,11 +609,11 @@ describe('setup-node', () => {
       await main.run();
 
       // assert
-      expect(logSpy).toHaveBeenCalledWith('LTS version is provided. For LTS versions `check-latest` will be automatically set to true')
-      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve the latest version from manifest...');
+      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve LTS alias from manifest...')
+      expect(dbgSpy).toHaveBeenCalledWith('Getting manifest from actions/node-versions@main')
+      expect(dbgSpy).not.toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
       expect(dbgSpy).toHaveBeenCalledWith(`LTS alias '*' for Node version 'lts/*'`)
       expect(dbgSpy).toHaveBeenCalledWith(`Found LTS release '14.0.0' for Node version 'lts/*'`)
-      expect(logSpy).toHaveBeenCalledWith("Resolved as '14.0.0'");
       expect(logSpy).toHaveBeenCalledWith(`Found in cache @ ${toolPath}`);
       expect(cnSpy).toHaveBeenCalledWith(`::add-path::${toolPath}/bin${osm.EOL}`);
     });
@@ -632,12 +637,12 @@ describe('setup-node', () => {
       await main.run();
 
       // assert
-      expect(logSpy).toHaveBeenCalledWith('LTS version is provided. For LTS versions `check-latest` will be automatically set to true')
-      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve the latest version from manifest...');
+      expect(logSpy).toHaveBeenCalledWith('Attempt to resolve LTS alias from manifest...')
+      expect(dbgSpy).toHaveBeenCalledWith('Getting manifest from actions/node-versions@main')
+      expect(dbgSpy).not.toHaveBeenCalledWith('No manifest cached, getting manifest from actions/node-versions@main')
       expect(dbgSpy).toHaveBeenCalledWith(`LTS alias '*' for Node version 'lts/*'`)
       expect(dbgSpy).toHaveBeenCalledWith(`Found LTS release '14.0.0' for Node version 'lts/*'`)
-      expect(logSpy).toHaveBeenCalledWith("Resolved as '14.0.0'");
-      expect(logSpy).toHaveBeenCalledWith("Attempting to download 14.0.0...");
+      expect(logSpy).toHaveBeenCalledWith("Attempting to download 14...");
       expect(logSpy).toHaveBeenCalledWith(`Acquiring 14.0.0 - ${os.arch} from ${expectedUrl}`);
       expect(logSpy).toHaveBeenCalledWith('Extracting ...');
       expect(logSpy).toHaveBeenCalledWith('Adding to the cache ...');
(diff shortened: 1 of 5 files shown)


src/installer.ts
```diff
@@ -183,6 +218,11 @@ async function getInfoFromManifest(
     auth,
     'main'
   );
+
+  if (isLts(versionSpec)) {
+    versionSpec = findLtsVersionFromManifest(versionSpec, stable, releases);
```
gordey4doronin, line 223: JFYI:

I tried to avoid fetching manifest multiple times, and incorporated the LTS resolving logic here. Since I set `check-latest = true` always, the resolving happens **before** toolcache code is executed, and the version is resolved at that point.

However, if we discuss and decide that `check-latest` should be automatically set to `true`, then I'll need to refactor code a little bit, and add additional manifest request before toolcache logic.

src/installer.ts
```diff
@@ -34,6 +34,11 @@ export async function getNode(
   let osPlat: string = os.platform();
   let osArch: string = translateArchToDistUrl(arch);
 
+  if (isLts(versionSpec)) {
+    core.warning('LTS version is provided. For LTS versions `check-latest` will be automatically set to true');
```
maxim-lobanov, line 38: No need to throw warning here because it produces a lot of noise in build results. `core.info` should be enough

src/installer.ts
```diff
@@ -170,6 +175,36 @@ export async function getNode(
   core.addPath(toolPath);
 }
 
+function isLts(versionSpec: string): boolean {
```
maxim-lobanov, line 178: nitpick: may be rename to `isLtsVersion`?

src/installer.ts
```diff
@@ -170,6 +175,36 @@ export async function getNode(
...
+  return versionSpec.startsWith('lts')
+}
+
+function findLtsVersionFromManifest(
+  versionSpec: string,
+  stable: boolean,
+  candidates: Array<tc.IToolRelease & { lts?: string }>
```
maxim-lobanov, line 185: Could you please create new type for candidates
```
type INodeRelease = tc.IToolRelease & { lts?: string }
```
I think it can be reused somewhere

src/installer.ts
```diff
@@ -170,6 +175,36 @@ export async function getNode(
...
+): string {
+  const alias = versionSpec.split('lts/')[1]?.toLowerCase();
+
+  if (!alias) {
+    throw new Error(`Unexpected LTS alias '${alias}' for Node version '${versionSpec}'`);
+  }
+
+  core.debug(`LTS alias '${alias}' for Node version '${versionSpec}'`);
+
+  const release = alias === '*'
```
maxim-lobanov, line 195: As I understand, we support the following syntax:
- `lts/*` - latest LTS
- `lts/<alias>` - specific LTS label

I would suggest to document it shortly via comment.

src/installer.ts
```diff
@@ -170,6 +175,36 @@ export async function getNode(
...
+   ? candidates.find(x => !!x.lts && x.stable === stable)
+   : candidates.find(x => x.lts?.toLowerCase() === alias && x.stable === stable);
+
+  if (!release) {
+    throw new Error(`Unable to find LTS release '${alias}' for Node version '${versionSpec}'.`);
+  }
+
+  core.debug(`Found LTS release '${release.version}' for Node version '${versionSpec}'`);
+
+  return release.version.split('.')[0];
```
maxim-lobanov, line 205: Why do we need to return major version only? What will happen if we will return exact version? 
I think `resolveVersionFromManifest` will handle it correctly.

src/installer.ts
maxim-lobanov: That is the good question. Our general politic is using cache as much as possible. 

If customer specifies `14.x` as input, we will check local cache firstly. In my opinion, `lts/fermium` input should work exactly the same way like `14.x` because it means the same.

May be @konradpabjan @bryanmacfarlane have different opinion.

maxim-lobanov: That is the good question. 
Our general policy is using local cache as much as possible.

If user specifies version input as `14.x`, we will check local cache firstly. In my opinion, `lts/fermium` input should work exactly the same way like `14.x` (be full alias) so we don't need to set check-latest to true.

konradpabjan: Agree

For `setup-java` we set `chcek-latest=false`:

https://github.com/actions/setup-java/blob/ad01d131cc39a1e467480c19bf28c5a4fade0edc/action.yml#L23-L25

I think we should be consistent across the board. One of the main justifications was to help speed up workflows. If the tool-cache is preferred than the majority of workflows will be ever so slightly faster (and that reduces COGS and infrastructure costs for us 😃 ). If something is not in the tool-cache than the action will reach out to externally try to find it. Sometimes there might be delays before the cache is updated with the latest version but i think it's minimal in the broader context of things. If someone absolutely needs the the latest versions always they can set `check-latest=true` themselves but in terms of optimizing for the most common use-case, I think `check-latest` should be `false`.

src/installer.ts
gordey4doronin: @maxim-lobanov Good point. I'll use `interface` then, to be more consistent with the existing code (which as far as I can see gives preference to interfaces). No objections?

```ts
interface INodeRelease extends tc.IToolRelease {
  lts?: string;
}
```

src/installer.ts
gordey4doronin: ✔️

src/installer.ts
gordey4doronin: ✔️

src/installer.ts
gordey4doronin: ✔️

src/installer.ts
gordey4doronin: OK, will change.
For my learning, what do you mean by "a lot of noise"?

This 👇 looks okayish to me 🤔 

Screenshot 2021-06-21 at 11 45 08

maxim-lobanov: Warning are also published in job summary that looks pretty annoying to me. (Like "something goes wrong" and I need to fix something). Especially, when customer runs job matrix, they will see a ton of such warnings.
image

src/installer.ts
maxim-lobanov: Looks good, thanks!

src/installer.ts
gordey4doronin: Thanks for the clarification! Now I see and agree, it's noisy 👍 

Screenshot 2021-06-21 at 14 24 18

gordey4doronin: I've fixed it for now. But I guess this code will go away anyway once we settle on `check-latest` usage.

src/installer.ts
gordey4doronin: I see, makes perfect sense.
My initial idea was to have closer behavior with `nvm`.
E.g. when you do `nvm use 14` it will use **any** existing 14th node on your machine.
But when you do `nvm use lts/fermium` it will always check `nodejs` website, and will use the **latest** 14th node.

But in the context of GitHub Action it's a little bit different. There is no "local machine", and basically on every run an action installs the latest possible node (whether it be from toolcache or from node website).

> Sometimes there might be delays before the cache is updated with the latest version but i think it's minimal in the broader context of things.

👆 I agree, then it's fine to use toolcache, and not set `check-latest`.
Especially considering this is a general policy. 👍 

**TL;DR** I initially thought within the bound of `nvm`, my oversight. In the context of GitHub Actions it makes sense to not set `check-latest=true`. I'll need to refactor the code a little bit (to fetch manifest separately), so please don't review yet. 😅

src/installer.ts
gordey4doronin: @maxim-lobanov Thinking again, maybe even `isLtsAlias`? Since it's actually checking whether it's an alias or not, instead of checking whether the version is actual LTS or not.

src/installer.ts
gordey4doronin: @maxim-lobanov @konradpabjan Remastered the whole logic. ✔️ 

Bonus: added `manifest` local variable, so if there was manifest fetched for `lts` check, there won't be another manifest fetch.

However, I didn't touch the existing (for smaller diff, easier review, and lower regression chance).
That means first call to `resolveVersionFromManifest` doesn't "cache" `manifest` as before, thus the second call to `getInfoFromManifest` will fetch `manifest` again. Hope that makes sense.

src/installer.ts
gordey4doronin: ✔️

src/installer.ts
```diff
@@ -170,20 +196,68 @@ export async function getNode(
...
+function resolveLtsAliasFromManifest(
+  versionSpec: string,
+  stable: boolean,
+  manifest: INodeRelease[]
+): string {
+  const alias = versionSpec.split('lts/')[1]?.toLowerCase();
+
+  if (!alias) {
+    throw new Error(
+      `Unexpected LTS alias '${alias}' for Node version '${versionSpec}'`
```
gordey4doronin, line 212: Personal thought:
Such line breaks caused by `"printWidth": 80`.
Increasing it to 100 would make code a little bit nicer.

CC @maxim-lobanov

__tests__/installer.test.ts
```diff
@@ -525,4 +544,252 @@ describe('setup-node', () => {
...
+      inputs.stable = 'true';
+
+      findSpy.mockImplementation(() => '');
+
+      // act
+      await main.run();
+
+      // assert
+      expect(logSpy).toHaveBeenCalledWith(
+        'Attempt to resolve LTS alias from manifest...'
```
gordey4doronin, line 753: Same `"printWidth": 80` issue here.

maxim-lobanov (approved):

maxim-lobanov:
@gordey4doronin , could you please also add e2e test. See example: https://github.com/actions/setup-node/blob/main/.github/workflows/versions.yml#L32
I think adding new job `lts-syntax` with `node-version: ["lts/*", "lts/Fermium"]` should be enough.

P.S. I believe we also need to update docs but we will take care about it in separate PR since we will rework docs in scope of https://github.com/actions/setup-node/pull/272

gordey4doronin:
Screenshot 2021-06-23 at 09 35 15


It seems some mocks are lacking for windows test. Will fix later.

gordey4doronin:
> could you please also add e2e test.

Absolutely. Good point. Will add too 👍

gordey4doronin:
* Fixed unit tests for windows (was just a path resolution problem)
* Added e2e tests for lts syntax
* Unfortunately had to remove `__tests__/verify-node.sh` step from it, since the check script doesn't allow to parse lts syntax. However, I it should be enough to have only the main step, it's running `actions/checkout@v2` and verifying lts alias resolution. If it couldn't resolve lts alias - it will fail. Profit.

Kikobeats:
this is a super great functionality that has been not implemented until now, even this package is used for a lot of maintainers.

@maxim-lobanov @konradpabjan can you merge this, please? 🙏

konradpabjan (approved):

dmitry-shibanov (approved):

__tests__/installer.test.ts
```diff
@@ -525,4 +548,244 @@ describe('setup-node', () => {
...
+        `LTS alias 'unknown' for Node version 'lts/unknown'`
+      );
+      expect(cnSpy).toHaveBeenCalledWith(
+        `::error::Unable to find LTS release 'unknown' for Node version 'lts/unknown'.${osm.EOL}`
+      );
+    });
+
+    it('fail if manifest is not available', async () => {
+      // arrange
+      os.platform = 'linux';
```
AlenaSviridenko, line 764: Could you please move this common setup to beforeAll hook of this describe?

src/installer.ts
```diff
@@ -170,20 +192,65 @@ export async function getNode(
...
```
AlenaSviridenko, line 213: Empty quotes in the error message looks a bit strange to me, maybe we should re-phrase to something like `LTS alias is empty for Node version '${versionSpec}'`

AlenaSviridenko (approved):

vsafonkin (approved):

__tests__/installer.test.ts
gordey4doronin: Makes sense 👍

I initially took `describe('check-latest flag'` as an example for my implementation, and it just duplicates everything right away.

But I agree, it would be nicer to move all common setup into a hook. 👍 I went ahead and did that for my new tests, without touching the existing tests of course.

Just to point out, `beforeAll` doesn't work in that case, because `inputs` and `os` are defined in the topmost `beforeEach` hook, and they're "new" in every single test case (`it`). So for my case I also had to use `beforeEach`.

src/installer.ts
gordey4doronin: Good catch, changed to something more generic.

gordey4doronin: @AlenaSviridenko Please let me know if that looks good to you. I decided to not be specific about alias being **empty**, but rather throw generic parse error.

AlenaSviridenko: I like how it looks now 🙂

AlenaSviridenko (approved):

maxim-lobanov:
@gordey4doronin  @Kikobeats , Merged PR. I will cut new version later today. Need to merge one more PR before that.

JimiC:
Good to see that it only took you about 2 years to finally implement "basic" functionality. Let's hope we, now, can use this action in our CI/CD.

P.S. Sorry if I sound cynical but 2 years is a loooong time for such a vital CI/CD component.

skjnldsv:
Thank you everyone that worked on this! Great addition! :tada: :hugs: :heart:

JimiC:
@skjnldsv What you don't know and that's why you reacted with 👎 is that I was the first to submit a PR with the exact same functionality 2 YEARS ago.

skjnldsv:
@JimiC sorry, I understand your frustration, but I don't see how your message helps anyone nor benefit the community in any sort! :confused: 

Have a great day! ☀️

JimiC:
> @JimiC sorry, I understand your frustration, but I don't see how your message helps anyone nor benefit the community in any sort! 😕

@skjnldsv It's just that when GH actions went live I was so excited about it and one of the first to try it out. So much that I even tried to contribute to the community and the community turned their back on me. So they made me from a GH fan to a GitLab fan.

> Have a great day! ☀️
You too. ☮️

mrlubos:
@gordey4doronin Can you help me to clarify my understanding? Can I at this point remove the `node-version` argument from the action setup and it will be able to infer the correct version from my `.nvmrc` file? If not, why? Would it be possible to add this functionality?

gordey4doronin:
@mrlubos Unfortunately you can't.👇 I wish that too. 🙏 

> Reading the `.nvmrc` file is not the goal of this PR.
> A separate PR shall be created for that


"Resolving aliases" and "reading version manager file" are two different technical tasks.
They're related to each other from business PoV.
But they're not related from code/implementation PoV.

My goal in the first place was to add LTS aliases resolving.

Now it's possible to use [this workaround](https://github.com/actions/setup-node/issues/32#issuecomment-539794249) if you have aliases in your `.nvmrc` file (which I do).

Reading the `.nvmrc` file is a standalone task, and was out of scope of my completely **voluntarily** PR.

For reading the file I personally up-voted https://github.com/actions/setup-node/issues/32 and https://github.com/actions/runner/issues/1180. 🙂 

Hope that makes sense.

mrlubos:
> @mrlubos Unfortunately you can't.👇 I wish that too. 🙏
> 
> > Reading the `.nvmrc` file is not the goal of this PR.
> > A separate PR shall be created for that
> 
> "Resolving aliases" and "reading version manager file" are two different technical tasks.
> They're related to each other from business PoV.
> But they're not related from code/implementation PoV.
> 
> My goal in the first place was to add LTS aliases resolving.
> 
> Now it's possible to use [this workaround](https://github.com/actions/setup-node/issues/32#issuecomment-539794249) if you have aliases in your `.nvmrc` file (which I do).
> 
> Reading the `.nvmrc` file is a standalone task, and was out of scope of my completely **voluntarily** PR.
> 
> For reading the file I personally up-voted #32 and [actions/runner#1180](https://github.com/actions/runner/issues/1180). 🙂
> 
> Hope that makes sense.

Thanks @gordey4doronin! I checked the issue you linked and what do you know, my vote is already there, too 😃 Have a nice day!

Issue #271, opened 2021-06-22
Support cache of dependencies in setup-node.


dmitry-shibanov:
# Context
`actions/setup-node` is the 2nd most popular action in GitHub Actions. A lot of customers use it in conjunction with [actions/cache](https://github.com/actions/cache) to speed up dependencies installation.  
See more examples on proper usage in [actions/cache documentation](https://github.com/actions/cache/blob/main/examples.md#node---npm).

# Goals & Anti-Goals
Integration of caching functionality into `actions/setup-node` action will bring the following benefits for action users:
- Decrease the entry threshold for using the cache for Node.js dependencies and simplify initial configuration
- Simplify YAML pipelines because no need additional steps to enable caching
- More users will use cache for Node.js so more customers will have fast builds!

We will add support for NPM and Yarn dependencies caching.  
As the first stage, we won't support custom locations for `package-lock.json`, `yarn.lock` files and action will work only when files are located in repository root. 

We don't pursue the goal to provide wide customization of caching in scope of `actions/setup-node` action. The purpose of this integration is covering ~90% of basic use-cases. If user needs flexible customization, we should advice them to use `actions/cache` directly.

dmitry-shibanov:
## Summarized tables.

### Express

| Package Manager (Cache Status) | Ubuntu    | Windows   | macOS    |
| ------------------------------ |:---------:|:---------:|:---------|
| npm (-)                        | 0:05      | 0:23      | 0:13     | 
| npm (+)                        | 0:05      | 0:19      | 0:10     | 
| yarn 1 (-)                     | 0:06      | 0:24      | 0:12     | 
| yarn 1 (+)                     | 0:05      | 0:21      | 0:10     | 
| yarn 2 (-)                     | 0:14      | 0:18      | 0:15     | 
| yarn 2 (+)                     | 0:12      | 0:16      | 0:13     |

### Nest

| Package Manager (Cache Status) | Ubuntu    | Windows   | macOS    |
| ------------------------------ |:---------:|:---------:|:---------|
| npm (-)                        | 1:02      | 1:40      | 1:41     | 
| npm (+)                        | 0:55      | 1:25      | 1:17     | 
| yarn 1 (-)                     | 0:43      | 3:44      | 5:27     |
| yarn 1 (+)                     | 0:26      | 3:21      | 4:01     |
| yarn 2 (-)                     | 2:01      | 2:30      | 3:01     | 
[...]

PR #272, opened 2021-06-22
Implementation of node's caching

dmitry-shibanov:
In scope of this pull request we add possibility to cache node dependencies. Add cache input parameter to actions/setup-node. For now, input will accept the following values:
 - npm - enable caching for npm dependencies
 - yarn - enable caching for yarn dependencies

Related issue: https://github.com/actions/setup-node/issues/271
### Description
Action will try to search package-lock.json or yarn.lock (npm 7.x supports yarn.lock files) files in the repository root and throw error if no one is found. The hash of found file will be used as cache key (the same approach like actions/cache recommends). The following key cache will be used ${{ runner.os }}-npm-${{ hashFiles('<package-lock-path>') }}
Action will cache global cache:
 - Npm (retrieved via `npm config get cache`)
 - Yarn 1 (retrieved via `yarn cache dir`)
 - Yarn 2 (retrieved via `yarn config get cacheFolder`)

#### Example of yml

- For npm
```yml
 uses: actions/setup-node@v2
  with:
    node-version: '14'
    cache: 'npm'
```

-  For yarn:
```yml
 uses: actions/setup-node@v2
  with:
    node-version: '14'
    cache: 'yarn'
```

 ADR: https://github.com/actions/setup-node/pull/266
 **Note:** 
 - It's not breaking change because it requires additional field and nothing changes by default.
 - License tests are falling. But we have changes to fix it. We've decided to stay with implementation, because approximately 50 licenses were changed. When changes are reviewed, we will add licenses to the existing pull request.
 
### What's done:
 - [x] - Implementation of node's caching
 - [x] - Adding tests for node's caching
 - [x] - Adding documentation
 - [x] - Update licenses


AlenaSviridenko (approved):

README.md
```diff
@@ -4,67 +4,69 @@
...
     node-version: '14'
-    check-latest: true
+    cache: 'yarn'
 - run: npm install
 - run: npm test
 ```
+Yarn caching handles both yarn versions: 1 or 2. 
 
-Matrix Testing:
+> At the moment, only `lock` files in the project root are supported.
```
bbugh, line 63: > At the moment, only `lock` files in the project root are supported.

Does this mean that `working-directory` will be ignored? That would be a deal breaker for mixed language mono repos.

maxim-lobanov: afaik, actions/setup-node doesn't support `working-directory` argument for now. `working-directory` can't be applied to actions steps (`uses:`).

In the first iteration, we have decided to keep things as simple as possible and we are looking for `package-lock.json` / `yarn.lock` only in repository root. We understand that it won't work for mono-repos (or any complex repos) so our recommendation is continue using [actions/cache](https://github.com/actions/cache/blob/main/examples.md#node---npm) directly for such use-cases for now.

We definitely will keep our eyes on feature requests to add supporting mono-repos but we wouldn't like to include it to the first iteration.

.github/workflows/e2e-cache.yml
```diff
@@ -0,0 +1,101 @@
+name: e2e-cache
+
+on:
+  pull_request:
+    paths-ignore:
+      - '**.md'
```
maxim-lobanov, line 6: nitpick: trailing spaces

.github/workflows/versions.yml
```diff
@@ -6,7 +6,7 @@ on:
       - '**.md'    
   push:    
     branches:
-      - main
+      - main
```
maxim-lobanov, line 9: please remove trailing spaces and revert this file

src/cache-restore.ts
```diff
@@ -0,0 +1,57 @@
...
+  const platform = process.env.RUNNER_OS;
+
+  const cachePath = await getCacheDirectoryPath(
+    packageManagerInfo,
+    packageManager
+  );
+  const lockFilePath = findLockFile(packageManagerInfo);
+  const fileHash = await glob.hashFiles(lockFilePath);
+
+  const primaryKey = `${platform}-${packageManager}-${fileHash}`;
```
maxim-lobanov, line 28: May be we can add `node-cache` prefix here?
`node-cache-${platform}-${packageManager}-${fileHash}`

src/cache-save.ts
```diff
@@ -0,0 +1,50 @@
+import * as core from '@actions/core';
+import * as cache from '@actions/cache';
+import {State} from './constants';
+import {getCacheDirectoryPath, getPackageManagerInfo} from './cache-utils';
+
+export async function run() {
+  const cacheLock = core.getInput('cache');
```
maxim-lobanov, line 7: Should we move this line under `try`?

maxim-lobanov (approved):

al-cheb (approved):

README.md
```diff
@@ -4,67 +4,69 @@
   <a href="https://github.com/actions/setup-node/actions?query=workflow%3Abuild-test"><img alt="build-test status" src="https://github.com/actions/setup-node/workflows/build-test/badge.svg"></a> <a href="https://github.com/actions/setup-node/actions?query=workflow%3Aversions"><img alt="versions status" src="https://github.com/actions/setup-node/workflows/versions/badge.svg"></a> <a href="https://github.com/actions/setup-node/actions?query=workflow%3Aproxy"><img alt="proxy status" src="https://github.com/actions/setup-node/workflows/proxy/badge.svg"></a> 
 </p>
 
-This action sets by node environment for use in actions by:
+This action provides the following functionality for GitHub Actions runners:
```
konradpabjan, line 7:
```suggestion
This action provides the following functionality for GitHub Actions users:
```

action.yml
```diff
@@ -19,6 +19,8 @@ inputs:
   token:
     description: Used to pull node distributions from node-versions.  Since there's a default, this is typically not supplied by the user.
     default: ${{ github.token }}
+  cache:
+    description: 'Used to specify package manager for caching in default directory. Supported values: npm, yarn'
```
konradpabjan, line 23:
```suggestion
    description: 'Used to specify a package manager for caching in the default directory. Supported values: npm, yarn'
```

konradpabjan (approved):

.github/workflows/e2e-cache.yml
dmitry-shibanov: Done. Thank you.

.github/workflows/versions.yml
dmitry-shibanov: Done. Thank you.

src/cache-restore.ts
dmitry-shibanov: Done. Thank you.

src/cache-save.ts
dmitry-shibanov: Done. Thank you.

README.md
dmitry-shibanov: Done. Thank you.

action.yml
dmitry-shibanov: Done. Thank you.

bryanmacfarlane (approved):
LGTM.  good job!

merceyz:
> The following key cache will be used ${{ runner.os }}-npm-${{ hashFiles('') }}

A bit late to the party but the cache for `yarn@>=2` is identical across all operating systems so that doesn't need to be part of the key

0-vortex:
Some high entropy questions:
- any particular reason for the decision that `shrinkwrap` is not a lockfile
- why is this action called `setup-node` if it only fully supports `yarn`
- why is it tagged `v2` if it's a breaking change, when `v2` even had a `v2-beta`

This PR also says it's only supporting files in the root of the module but that just seems like cutting corners on customising the cache key, providing opinionated defaults without a way to configure them is simply pushing this PR's opinions on people or forcing them to compete in implementation. 

Currently, the only way forward is to completely change the way we use `npm` (slash migrate to yarn) or switch all the workflows to `v2-beta` in order to avoid various tools picking the change up and suggesting action update.

jacobwgillespie:
> why is it tagged `v2` if it's a breaking change

@0-vortex see above, specifically this line:

> It's not breaking change because it requires additional field and nothing changes by default.

This PR adds a new feature where if you were previously using `actions/setup-node` and `actions/cache`, you might be able to remove `actions/cache` by adding the new `cache` key here. But you don't have to, you can continue using your current config, this PR doesn't change existing behavior.

0-vortex:
> > why is it tagged `v2` if it's a breaking change
> 
> @0-vortex see above, specifically this line:
> 
> > It's not breaking change because it requires additional field and nothing changes by default.
> 
> This PR adds a new feature where if you were previously using `actions/setup-node` and `actions/cache`, you might be able to remove `actions/cache` by adding the new `cache` key here. But you don't have to, you can continue using your current config, this PR doesn't change existing behavior.

Thank you for the reply but it is not a clarification, sorry if the comment felt rushed, have spent a decent amount of hours toying with this implementation without getting positive results. As a precaution I have read all the comment history and reviewed the code changes twice before posting the questions. 

Some references:
- https://github.com/open-sauced/open-sauced/runs/2986322631#step:3:16
- https://github.com/akv-platform/setup-node/blob/cfe70f9fe8ca345bad2e02a88695dd317e6172c0/src/cache-utils.ts#L15
- https://docs.npmjs.com/cli/v7/commands/npm-shrinkwrap
- https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#package-lockjson-vs-npm-shrinkwrapjson

Motivation here is to fix/implement cache for a number of failing repositories while promoting best practices.

maxim-lobanov:
@0-vortex , We don't pursue the goal to provide wide customization of caching in scope of `actions/setup-node` action. The purpose of this integration is covering ~90% of basic use-cases. If user needs flexible customization, they should continue to use [actions/cache](https://github.com/actions/cache).

Currently, this integration covers basic use-cases for both npm and yarn:
- https://github.com/actions/cache/blob/main/examples.md#node---npm
- https://github.com/actions/cache/blob/main/examples.md#node---yarn

As @jacobwgillespie fairly mentioned above, 
> This PR adds a new feature where if you were previously using actions/setup-node and actions/cache, you might be able to remove actions/cache by adding the new cache key here. But you don't have to, you can continue using your current config, this PR doesn't change existing behavior.

New functionality is added under additional flag `cache`. if you don't add new flag - everything will work as previously. So this PR is not breaking change.

Your suggestion about `shrinkwrap` file is fair enough and it would be helpful if you create feature request issue for it?
Until it is implemented, I suggest continue using [actions/cache](https://github.com/actions/cache) that provides more customized configuration.

0-vortex:
Can you share where you get those stats from? Shrinkwrap is a standard production practice, don't understand how you can call it a suggestion.

Thompson1985 (approved):

Thompson1985 (approved):

PR #277, opened 2021-06-30
Update readme to describe new NVM LTS syntax

maxim-lobanov:
Describe new NVM version syntax. Also bump versions in tests and docs


README.md
```diff
@@ -26,16 +26,24 @@ steps:
 - run: npm test
 ```
 
-The `node-version` input is optional. If not supplied, the node version from PATH will be used. However, this action will still register problem matchers and support auth features. So setting up the node environment is still a valid scenario without downloading and caching versions.
+The `node-version` input is optional. If not supplied, the node version from PATH will be used. However, It is recommended to always specify Node.js version and don't rely on system one.
```
AlenaSviridenko, line 29:
```suggestion
The `node-version` input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.  
```

AlenaSviridenko (approved):

gordey4doronin (commented):
👏

ONFU:
VALKHYLMHER VALENTIME VALKIRYE VENDETTA HAVED CHOSING GOD DISTURB FROM  MAIN LESS AERA SOVIETIC AERA FROM NEW RESPOWND HAVED CHANGING HALES OUT GATE OF UTOPIA

PR #278, opened 2021-06-30
Add pnpm caching support

jacobwgillespie:
This PR extends the excellent caching work at #272 with support for pnpm. 

The global module store is stored at `~/.pnpm-store` by default, unless otherwise configured. `pnpm get store` returns the custom store path, or the string `undefined` if no custom store is configured, so this PR adds a `defaultCacheFolder` option that is used as a fallback if the `getCacheFolderCommand` does not return a path. Note that the cache directory path is sent ultimately to `@actions/glob`, and that module handles the expansion of `~/` to the current home directory, so it's okay that the path is not expanded here.

Note: I've opened [a PR on `pnpm/pnpm`](https://github.com/pnpm/pnpm/pull/3571) to add a new command that returns the current module store directory, including the default if no custom path is set, but even if that PR is accepted, I believe we'd still need to support the `defaultCacheFolder` fallback here as current and past versions of pnpm would not support the new command.

I've added tests and Actions workflow jobs to match the existing npm and yarn 1 / 2 tests.

### DIFF
--- a/.github/workflows/e2e-cache.yml
+++ b/.github/workflows/e2e-cache.yml
@@ -35,6 +35,39 @@ jobs:
         run: __tests__/verify-node.sh "${{ matrix.node-version }}"
         shell: bash
 
+  node-pnpm-depencies-caching:
+    name: Test pnpm (Node ${{ matrix.node-version}}, ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        node-version: [12, 14, 16]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2
+        with:
+          version: 6.9.0
+      - name: Generate pnpm file
+        run: pnpm install
+      - name: Remove dependencies
+        shell: pwsh
+        run: Remove-Item node_modules -Force -Recurse
+      - name: Clean global cache
+        run: rm -rf ~/.pnpm-store
+        shell: bash
+      - name: Setup Node
+        uses: ./
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'pnpm'
+      - name: Install dependencies
+        run: pnpm install
+      - name: Verify node and pnpm
+        run: __tests__/verify-node.sh "${{ matrix.node-version }}"
+        shell: bash
+
   node-yarn1-depencies-caching:
     name: Test yarn 1 (Node ${{ matrix.node-version}}, ${{ matrix.os }})
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/e2e-cache.yml
+++ b/.github/workflows/e2e-cache.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v2
         with:
-          version: 6.9.0
+          version: 6.10.0
       - name: Generate pnpm file
         run: pnpm install
       - name: Remove dependencies
--- a/README.md
+++ b/README.md
@@ -55,18 +55,20 @@ steps:
 - run: npm test
 ```
 
-**Caching pnpm dependencies:**
+**Caching pnpm (v6.10+) dependencies:**
 ```yaml
 # This workflow uses actions that are not certified by GitHub.
 # They are provided by a third-party and are governed by
 # separate terms of service, privacy policy, and support
 # documentation.
 
+# NOTE: pnpm caching support requires pnpm version >= 6.10.0
+
 steps:
 - uses: actions/checkout@v2
 - uses: pnpm/action-setup@646cdf48217256a3d0b80361c5a50727664284f2
   with:
-    version: 6.9.0
+    version: 6.10.0
 - uses: actions/setup-node@v2
   with:
     node-version: '14'
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 This action provides the following functionality for GitHub Actions users:
 
 - Optionally downloading and caching distribution of the requested Node.js version, and adding it to the PATH
-- Optionally caching npm/pnpm/yarn dependencies
+- Optionally caching npm/yarn/pnpm dependencies
 - Registering problem matchers for error output
 - Configuring authentication for GPR or npm
 
@@ -41,7 +41,7 @@ nvm lts syntax: `lts/erbium`, `lts/fermium`, `lts/*`
 
 ### Caching packages dependencies
 
-The action has a built-in functionality for caching and restoring npm/yarn dependencies. Supported package managers are `npm`, `pnpm`, `yarn`. The `cache` input is optional, and caching is turned off by default.
+The action has a built-in functionality for caching and restoring npm/yarn dependencies. Supported package managers are `npm`, `yarn`, `pnpm`. The `cache` input is optional, and caching is turned off by default.
 
 **Caching npm dependencies:**
 ```yaml
@@ -54,6 +54,19 @@ steps:
 - run: npm install
 - run: npm test
 ```
+
+**Caching yarn dependencies:**
+```yaml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-node@v2
+  with:
+    node-version: '14'
+    cache: 'yarn'
+- run: yarn install
+- run: yarn test
+```
+Yarn caching handles both yarn versions: 1 or 2.
 
 **Caching pnpm (v6.10+) dependencies:**
 ```yaml
@@ -76,19 +89,6 @@ steps:
 - run: pnpm install
 - run: pnpm test
 ```
-
-**Caching yarn dependencies:**
-```yaml
-steps:
-- uses: actions/checkout@v2
-- uses: actions/setup-node@v2
-  with:
-    node-version: '14'
-    cache: 'yarn'
-- run: yarn install
-- run: yarn test
-```
-Yarn caching handles both yarn versions: 1 or 2. 
 
 > At the moment, only `lock` files in the project root are supported.
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 This action provides the following functionality for GitHub Actions users:
 
 - Optionally downloading and caching distribution of the requested Node.js version, and adding it to the PATH
-- Optionally caching npm/yarn dependencies
+- Optionally caching npm/pnpm/yarn dependencies
 - Registering problem matchers for error output
 - Configuring authentication for GPR or npm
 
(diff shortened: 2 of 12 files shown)


maxim-lobanov:
Hey @jacobwgillespie , thank you for your contribution, Your changes look awesome!
Please give us some time to review and validate it properly.

Did you have a chance to validate it on any real project? You can reference your fork branch via `uses: jacobwgillespie/setup-node@cache-pnpm` in your YAML file for e2e validation.

jacobwgillespie:
@maxim-lobanov no problem, take your time, and let me know if there's anything I can do to help as you're reviewing!

I did validate it on some internal projects - if you'd like a public example, I just updated the [`styled-icons` repo](https://github.com/styled-icons/styled-icons/runs/3058657720) to use the forked action, you can see an example of it loading the pnpm cache at that link.

src/cache-utils.ts
```diff
@@ -70,9 +78,16 @@ export const getCacheDirectoryPath = async (
...
-  );
+  let stdOut = await getCommandOutput(packageManagerInfo.getCacheFolderCommand);
+
+  // pnpm returns 'undefined' if no custom store path is set
+  if (stdOut === 'undefined') {
+    stdOut = '';
+  }
+
+  if (!stdOut && packageManagerInfo.defaultCacheFolder) {
+    stdOut = packageManagerInfo.defaultCacheFolder;
```
maxim-lobanov, line 89: I think we should extend this function a bit and add validation check that retrieved cache path exists on disk.

The `toolkit/cache` package doesn't handle this situation properly and if cache path doesn't exist on disk, it won't fail. It will create malformed archive with size 22 B silently and then fail to restore such cache. We had similar issue with Yarn recently: https://github.com/actions/setup-node/issues/281 .

Since we fallback to `~/.pnpm-store` in case of issue, it would be great to add such validation to check if this folder exists and fail earlier if it doesn't exist.
I suggest something like this:
```typescript
if (!stdOut && packageManagerInfo.defaultCacheFolder) {
  stdOut = packageManagerInfo.defaultCacheFolder;
}

if (!stdOut) {
  throw new Error(`Could not get cache folder path for ${packageManager}`);
}
core.debug(`${packageManager} path is ${stdOut}`);

if (!fs.existsSync(stdOut)) {
  throw new Error(`Cache folder path is retrieved for ${packageManager} but doesn't exist on disk: ${stdOut}`);
}

return stdOut;
```

The single problem that `fs.exists` won't handle `~` in path properly but we can replace `defaultCacheFolder: '~/.pnpm-store'` with `defaultCacheFolder: path.join(os.homedir(), '.pnpm-store')`
Please let me know if you see any alternative options

maxim-lobanov (commented):
Everything looks good to me. I have left minor comment to add additional validation.

I will ping more engineers to review it from our side and also we will do basic validation from our side.

__tests__/cache-save.test.ts
```diff
@@ -150,6 +152,23 @@ describe('run', () => {
       );
       expect(setFailedSpy).not.toHaveBeenCalled();
     });
+
+    it('should not save cache for npm', async () => {
```
konradpabjan, line 156:
```suggestion
    it('should not save cache for pnpm', async () => {
```

I think it should be this?

README.md
```diff
@@ -55,6 +55,21 @@ steps:
 - run: npm test
 ```
 
+**Caching pnpm dependencies:**
+```yaml
+steps:
+- uses: actions/checkout@v2
+- uses: pnpm/action-setup@v2
```
konradpabjan, line 62: Since this is an third party action not supported officially by us can you add this disclaimer at the top of the YAML snippet

```
# This workflow uses actions that are not certified by GitHub.
# They are provided by a third-party and are governed by
# separate terms of service, privacy policy, and support
# documentation.
```

Could you also pin to a specific SHA version instead of `v2` for the `pnpm` action. This is more for security so that in-case something is compromised with the third-party action than potential impact is reduced.

Here is a similar example that can be used as a reference: https://docs.github.com/en/actions/guides/building-and-testing-python#publishing-to-package-registries

src/cache-utils.ts
jacobwgillespie: This makes sense to me, I have updated with your suggestion 👍

__tests__/cache-save.test.ts
jacobwgillespie: Good catch, fixed!

README.md
jacobwgillespie: Updated, I've added the disclaimer and pinned to the latest commit on the `v2` branch 👍

maxim-lobanov (approved):

src/cache-utils.ts
maxim-lobanov: Thanks, but some tests fail. I think we should also mock `fs.existsSync` to return `true` in tests.

jacobwgillespie: I think the test failures are legitimate, this function is used to compute the cache path when _restoring_ the cache as well. Do you have any preferences on how to proceed? I think we only want to throw an error on the save path, not on the restore path.

jacobwgillespie: I moved the `if (!fs.existsSync(cachePath))` check to `cache-save.ts`, that way it only throws that error when attempting to save the cache.

konradpabjan (approved):

src/cache-utils.ts
maxim-lobanov: Personally, I would prefer to keep this check in `getCacheDirectoryPath` function. When we calculate cache path from tool - we check that this path is correct.
Looks like some tests still fail. Have you tried to run it locally? Does it work for you? I guess we still need to mock `fs.existSync` invocation.

jacobwgillespie: @maxim-lobanov I don't believe the error can stay in `getCacheDirectoryPath` because of this:

https://github.com/actions/setup-node/blob/38d90ce44d5275ad62cc48384b3d8a58c500bb5f/src/cache-restore.ts#L21-L33

The directory name returned by `getCacheDirectoryPath` is used to compute the cache path _before_ it's restored, so `fs.existsSync` would always fail, as the cache is not restored at that point. That's why I think the check only applies to `cache-save.ts`, as it's only the save path that needs to check the existence of the directory.

I'll work on mocking `fs.existsSync` for the cache save tests...

jacobwgillespie: Okay, `fs.existsSync` is mocked, all tests are passing for me locally ✅

maxim-lobanov: Hm, auth tests fail in CI for some reason. May be mocks are not clean up properly? We can help with investigation.

Also, I would like to discuss one more issue: 
`defaultCacheFolder: path.join(os.homedir(), '.pnpm-store')` doesn't work for Windows (e2e CI fails on this platform).
I have tried to experiment and looks like pnpm uses different cache dir for different disk drives. The following commands:
```
New-Item -ItemType Directory -Path "C:\Temp2"
cd "C:\Temp2"
pnpm store path
          
New-Item -ItemType Directory -Path "D:\Temp2"
cd "D:\Temp2"
pnpm store path
```
produces the following output:
```
C:\Users\runneradmin\.pnpm-store\v3
D:\.pnpm-store\v3
```

Looks like cache path on Windows depends on current directory so we can't simply set `defaultCacheFolder`.

I wonder if we can just rely on `pnpm store path` command introduced by you without fallback path and note in README that caching for pnpm works for pnpm 6.10+
We don't have much experience with pnpm so we would like to hear your opinion.

jacobwgillespie: I'm comfortable with relying on `pnpm store path`, now that it's released. I'll push that change now, so that we can see if tests will pass. (I'll also set the `fs` mock to unmock after the cache save tests)

An alternative or addition would be to import the [`@pnpm/store-path` package](https://www.npmjs.com/package/@pnpm/store-path), which allows you to programatically resolve the store path. I think it would be best used as a fallback for if `pnpm store path` failed. However, I assume that you'd prefer to not include another dependency in this action, and would prefer to avoid package-manager-specific integrations. But let me know if I'm incorrect about that.

.github/workflows/e2e-cache.yml
```diff
@@ -35,6 +35,39 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        node-version: [12, 14, 16]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2
+        with:
+          version: 6.9.0
```
maxim-lobanov, line 51:
```suggestion
          version: 6.10.0
```

src/cache-utils.ts
maxim-lobanov: I like the current approach and I think we should proceed with it for now. We can revisit in future if users really want to use older versions of pnpm but I wouldn't add new dependencies here.

Looks like all tests are green, thank you!

maxim-lobanov:
@bryanmacfarlane , could you please review and approve this PR when you have a chance? If you don't have objections - we can release new action version next week with these changes.

action.yml
```diff
@@ -20,7 +20,7 @@ inputs:
     description: Used to pull node distributions from node-versions.  Since there's a default, this is typically not supplied by the user.
     default: ${{ github.token }}
   cache:
-    description: 'Used to specify a package manager for caching in the default directory. Supported values: npm, yarn'
+    description: 'Used to specify a package manager for caching in the default directory. Supported values: npm, pnpm, yarn'
```
bryanmacfarlane, line 23: nit: since we're adding this one on, can we swap the order and add pnpm after yarn.  interesting question whether it's alphabetical or by usage.  not sure we can answer that but anecdotally, I think it should be npm, yarn, pnpm.  as I said a nit and not passionate about it.

jacobwgillespie: Thoughts on this as it relates to the README? I went with alphabetical order there as well.

bryanmacfarlane: My 2 cents is just add to the list in those two instances and then 👍 on changes.  Thanks so much for this valuable addition.

jacobwgillespie: Perfect, I have updated the order in `action.yml` and the README to npm, yarn, pnpm 👍

bryanmacfarlane (approved):

dmitry-shibanov (approved):

maxim-lobanov:
Hello @jacobwgillespie , just an update.
We have got all necessary approvals for this PR and finished with validation from our side.

We will be ready to cut new action release at the begin of next week.

jacobwgillespie:
Awesome! 🎉 Thank you so much, excited to see this released!

MaksimZhukov (approved):

MaksimZhukov:
Hello @jacobwgillespie! 
We have released `actions/setup-node@v2.3.0` and updated the `v2` tag. Thank you for the contribution!

Would you like to create a PR in the [github/docs](https://github.com/github/docs) repository to update GitHub Docs in accordance with your changes?

jacobwgillespie:
@MaksimZhukov perfect, thanks for the merge and release!

I've created a docs PR here: https://github.com/github/docs/pull/8389

Thompson1985 (approved):

PR #279, opened 2021-07-01
feat: add support for `current` and `latest` aliases

### LINKED ISSUE #257: Support `latest` as a version alias for the latest Node version
#26 discusses supporting several types of aliases. That discussion has been going on for a long time. It seems the main complications revolve around how to deal with `lts`.

For those of us who don't care about `lts` and just want `latest` so that we can test that our work works on latest Node automatically, could we please get support for `latest` first and let #26 focus on the subtleties of `lts` and how to implement that? It seems like these two aliases do not need to be grouped into a single feature release, in fact, it looks, like implementing them separately would simplify things and potentially make it possible to provide `latest` soon.

Divlo:
Fixes #257 

The goal of this PR is to support `current` as alias for the `node-version` array.
I thought that `current` is a better name than `latest` because the actual name from Node.js is `current` but we can support both aliases if we really want to.

Currently, this is a Work In Progress, I don't know much of this codebase, so I would appreciate some help.
I'm new to the world of developing GitHub Actions, I always use them but never develop them 😆, I saw that you can test if it actually works with the tool [nektos/act](https://github.com/nektos/act).
So I ran this command : `act -j current-syntax` and it seems to work, but it is not caching the Node.js version yet.

**EDIT:** I also added the `latest` alias.


Divlo:
cc @maxim-lobanov

TomasHubelbauer:
I'd like to request that 'latest' also be supported, however even without it this would be great to have! Thank you for making the PR

Divlo:
> I'd like to request that 'latest' also be supported, however even without it this would be great to have! Thank you for making the PR

Done! :+1: 
I also added the `latest` alias in addition to the `current` one.

Tungbengok:
Dowload ?

Divlo:
From https://github.com/actions/setup-node/pull/278#issuecomment-881615296
> We will be ready to cut new action release at the begin of next week.

Could you review this PR and maybe include it in the next release if possible and everything is good to go, please ? :smile: @maxim-lobanov

Artek199 (approved):

Kikobeats:
any chance to merge this?

demurgos:
Hi,
It's been two months: is there any news about this PR?

Divlo:
Thanks for your interest. @Kikobeats and @demurgos
 
I'm happy to continue to work on this PR if it's needed.
The main problem currently is that it is not caching the Node.js version, it's downloading again Node.js even if the current release hasn't changed.
That means it's slower than it should be, at least it works. :smile:
But to set up caching, I would need some help from the maintainers.

Divlo:
> Hello @Divlo. Thank you for your pull request. Could please attach link to the documentation for the `current` alias ? If I understood well, it should install the latest (lts/non lts) version ? My main concern is about that the latest alias is more obvious.

Updated README and resolved the conflicts. :+1: 
Yes, it installs the latest Node.js version (for example currently it is v17.3.1).
The `latest` alias might be more obvious but the real name is `current` as you can see on the [Node.js website](https://nodejs.org/).

image

aivus:
Latest is also used in the dists:
https://nodejs.org/dist/latest/

dominykas:
Just a memory dump of prior art:

- `current` is the official terminology as per https://nodejs.org/en/about/releases/
- `nvm` does not support `current` or `latest`, but it does support `node` (and a deprecated alias for that: `stable`).
- Travis CI (reference CI for a very long time in the Node.js ecosystem) does support `latest` (and pulls in the "current (stable) release"), but I think it does that because it installs `node` via nvm when the keyword is unsupported, rather than `latest` being explicitly supported (if I remember correctly from the research I did for https://github.com/pkgjs/detect-node-support)
- https://github.com/pkgjs/nv (reference package from https://github.com/nodejs/package-maintenance WG) supports `current` (alias meaning the same as `node` from `nvm`)

I'd argue `latest` should not be supported explicitly.

jehon:
Could it be that the two would exists ? "current" and "latest" being synonyms ?

dominykas:
My thinking is that "latest" can be misleading, esp. in the context of there being "latest in a release line", "latest LTS", etc. I'd rather have `latest` not work and error out, so that people do not pick that up and try to use it elsewhere, eventually leading to further confusion.

I could be wrong in my thinking, of course. I'll try to dig out some more discussion threads about this for reference.

dominykas:
I guess a lot of threads can be followed from https://github.com/nodejs/package-maintenance/issues/236

I do have to admit, that I have an issue open on `nvm` side to support `latest` myself, but ideally these keywords should all be converging...

Divlo:
@dmitry-shibanov Friendly ping. :smile:

What is needed to merge this PR?

milleniait:
Join the All New Cross chain project Dogelana $DGLN.

Doge and Solana introduces the best and fastest Network on Blockchain.
Join us to partake in our IEO/ICO/NFT GIVEAWAY

https://t.me/+ZtYF6PzehrMxYmE0

milleniait:
Join the All New Cross chain project Dogelana $DGLN.

Doge and Solana introduces the best and fastest Network on Blockchain.
Join us to partake in our IEO/ICO/NFT GIVEAWAY

https://t.me/+ZtYF6PzehrMxYmE0

bluelovers:
> Just a memory dump of prior art:
> 
> * `current` is the official terminology as per https://nodejs.org/en/about/releases/
> * `nvm` does not support `current` or `latest`, but it does support `node` (and a deprecated alias for that: `stable`).
> * Travis CI (reference CI for a very long time in the Node.js ecosystem) does support `latest` (and pulls in the "current (stable) release"), but I think it does that because it installs `node` via nvm when the keyword is unsupported, rather than `latest` being explicitly supported (if I remember correctly from the research I did for https://github.com/pkgjs/detect-node-support)
> * https://github.com/pkgjs/nv (reference package from https://github.com/nodejs/package-maintenance WG) supports `current` (alias meaning the same as `node` from `nvm`)
> 
> I'd argue `latest` should not be supported explicitly.

so does actions/setup-node support `node` too?

Divlo:
@bluelovers Does `node` alias is about the latest current Node.js released version (e.g: currently it is v17.5.0)?
If yes, that would solve this issue, but it just adds another naming...
What naming should we choose: `current` or `latest` or `node` or all the three together?

bluelovers:
> @bluelovers Does `node` alias is about the latest current Node.js released version (e.g: currently it is v17.5.0)? If yes, that would solve this issue, but it just adds another naming... What naming should we choose: `current` or `latest` or `node` or all the three together?

nvm , travis ci use `node` , and actions/setup-node support nvmrc, so i think `node` should be support

Divlo:
This PR has no recent activity, and probably won't be merged for a long time and as I don't want/have time to do the job for this PR, I decided to close it.

Feel free to open another PR yourself based on this one if you want this to be implemented.

Issue #282, opened 2021-07-02
Accept a working-directory as an input argument


Nefcanto:
I'm trying to setup node for my GitHub Action. But I'm not working in the default directory.

I clone a repo in `dir_one/rep_one` and then clone another repo in `dir_two/rep_two` and copy a couple of files from the first repo to the second repo.

Then I want to setup Node and `npm install` on the second repo. 

`working-directory` doesn't work with `uses` and I also tried to `- run: cd dir_two/repo_two`, but again build fails because it's running in its folder.

I receive this error:

> Error: Dependencies lock file is not found in /home/runner/work/Site/Site. Supported file patterns: package-lock.json,yarn.lock

Please create an input argument to accept path where node should search for `package.json` file.

karlosos:
I have a similar problem. My monorepo has the following structure:

```
my-repo/
├─ frontend/
│  ├─ package.json
│  ├─ package-lock.json
│  ├─ src
├─ backend/
```

Is there a possibility of using `actions/setup-node@v2` inside `./frontend` directory?

I've tried following workflow:

```
# This workflow will do a clean install of node dependencies, cache/restore them, build the source code and run tests across different versions of node
# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions

name: Node.js CI

on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]

defaults:
  run:
    working-directory: frontend

jobs:
  build:

    runs-on: ubuntu-latest

    strategy:
      matrix:
        node-version: [12.x, 14.x, 16.x]
        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/

    steps:
    - uses: actions/checkout@v2
    - name: Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v2
      with:
        node-version: ${{ matrix.node-version }}
        cache: 'npm'
    - run: npm ci
    - run: npm run build --if-present
    - run: npm test
```

With similar error:

> Error: Dependencies lock file is not found in /home/runner/work/django_chatroom/django_chatroom. Supported file patterns: package-lock.json,yarn.lock

![image](https://user-images.githubusercontent.com/3882385/124731395-788f6180-df12-11eb-9458-76442b7a560b.png)

dbaker-aa:
@Nefcanto , @karlosos 

[...]

Issue #275, opened 2021-06-29
Support cache dependencies for cases where package lock files are located in subfolders


bbugh:
Hi! 👋

We have a mixed-language monorepo with nothing in the base folder except some CI info. We need to be able to specify `working-directory` input like other scripts in order to use this fantastic action to build our node subfolders.

Something like:

```yaml
- uses: actions/setup-node@v2
  with:
    node-version: '14'
    working-directory: some-client
```

Thanks!

maxim-lobanov:
Hello @bbugh , I don't think that you need `working-directory` input in setup-node task. This task just installs specified Node.js version and do nothing with working directory.

If your `package.json` / project are located in the separate folder, I think you need to specify `working-directory` for next steps:
```yml
- uses: actions/setup-node@v2
  with:
    node-version: '14'
    
- name: Install dependencies and build project
  working-directory: some-client
  run: |
    npm install
    npm run build
    npm run test
```

bbugh:
Hi again! 👋 On [the caching PR](https://github.com/actions/setup-node/pull/272), I had asked about caching working for mono repos, and you had replied saying that's not supported yet. I created this issue to request that support. 😉 

Since GitHub actions has poor support for yaml and no reuse between steps, we have a lot of repetitive yarn tasks for caching and running. Having support built into this will be super helpful, and as we have a multi-language mono repo, we need to be able to run and cache them separately.

In particular, Yarn 2's cache folder will be different for each project, as the default is `<project>/.yarn` and in the case of zero-installs would be committed to source control. In that cache, a `working-directory` would be used to successfully run yarn in the correct location with the correct cache folder. Or a `cache-root` option, or something.

maxim-lobanov:
@bbugh caught it. I have renamed the issue title a bit to be more clear. Hope it works for you!

bbugh:
I'm always reminding people about XY problems and here I did one myself. Thanks!

maxim-lobanov:
Hello everyone!

This feature request will be implemented soon. I have a small ADR under review https://github.com/actions/setup-node/pull/299 .
Proposed approach is a bit different from proposed in this issue - we would like allow specifying full path (directory path + filename) to lock file instead of `working_directory` (directory path only). We believe this approach will be a bit more generic and unblock more use-cases.

dmitry-shibanov:
[...]

PR #299, opened 2021-07-13
ADR for support caching for mono repos and repositories with complex structure

maxim-lobanov:
Related feature requests:
- https://github.com/actions/setup-node/issues/275
- https://github.com/actions/setup-node/issues/282

### DIFF
--- a/docs/adrs/0001-support-caching-deps-for-monorepos.md
+++ b/docs/adrs/0001-support-caching-deps-for-monorepos.md
@@ -1,0 +1,48 @@
+# 0. Support caching dependencies for mono repos
+Date: 2021-07-13
+
+Status: Proposed
+
+## Context
+Currently, `actions/setup-node` supports caching dependencies for Npm and Yarn package managers.  
+For the first iteration, we have decided to not support cases where `package-lock.json` / `yarn.lock` are located outside of repository root.  
+Current implementation searches the following file patterns in the repository root: `package-lock.json`, `yarn.lock` (in order of resolving priorities)
+
+Obviosly, it made build-in caching unusable for mono-repos and repos with complex structure.  
+We would like to revisit this decision and add customization for dependencies lock file location.
+
+## Proposal
+We have the following options:
+1. Allow to specify  directory where `package-lock.json` or `yarn.lock` are located
+2. Allow to specify path to the dependencies lock file (including directory path and filename)
+
+The second option looks more generic because it allows to:
+- specify multiple dependencies files using file patterns like `**/package-lock.json` ([one of recommended approaches in actions/cache](https://github.com/actions/cache/blob/main/examples.md#macos-and-ubuntu))
+- specify custom dependencies files like `src/npm-shrinkwrap.json`
+- change default resolving priority if both `yarn.lock` and `package-lock.json` exist in repository
+
+## Decision
+
+Add `deps-lock-file` input that will accept path (relative to repository root) to dependencies lock file.  
+If path provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/packages.lock.json') }}` YAML construction does.  
+The hash of privided matched files will be used as a part of cache key.
+
+Yaml examples:
+```yml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-node@v2
+  with:
+    node-version: 14
+    cache: npm
+    deps-lock-file: 'sub-project/package-lock.json'
+```
+```yml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-node@v2
+  with:
+    node-version: 14
+    cache: yarn
+    deps-lock-file: 'sub-project/**/yarn.lock'
+```
--- a/docs/adrs/0001-support-caching-deps-for-monorepos.md
+++ b/docs/adrs/0001-support-caching-deps-for-monorepos.md
@@ -8,7 +8,7 @@ Currently, `actions/setup-node` supports caching dependencies for Npm and Yarn p
 For the first iteration, we have decided to not support cases where `package-lock.json` / `yarn.lock` are located outside of repository root.  
 Current implementation searches the following file patterns in the repository root: `package-lock.json`, `yarn.lock` (in order of resolving priorities)
 
-Obviosly, it made build-in caching unusable for mono-repos and repos with complex structure.  
+Obviously, it made build-in caching unusable for mono-repos and repos with complex structure.  
 We would like to revisit this decision and add customization for dependencies lock file location.
 
 ## Proposal
@@ -24,8 +24,8 @@ The second option looks more generic because it allows to:
 ## Decision
 
 Add `deps-lock-file` input that will accept path (relative to repository root) to dependencies lock file.  
-If path provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/packages.lock.json') }}` YAML construction does.  
-The hash of privided matched files will be used as a part of cache key.
+If provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/packages.lock.json') }}` YAML construction does.  
+The hash of provided matched files will be used as a part of cache key.
 
 Yaml examples:
 ```yml
--- a/docs/adrs/0001-support-caching-deps-for-monorepos.md
+++ b/docs/adrs/0001-support-caching-deps-for-monorepos.md
@@ -23,7 +23,7 @@ The second option looks more generic because it allows to:
 
 ## Decision
 
-Add `package-lock-file` input that will accept path (relative to repository root) to dependencies lock file.  
+Add `cache-dependency-path` input that will accept path (relative to repository root) to dependencies lock file.  
 If provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/package-lock.json') }}` YAML construction does.  
 The hash of provided matched files will be used as a part of cache key.
 
@@ -35,7 +35,7 @@ steps:
   with:
     node-version: 14
     cache: npm
-    package-lock-file: 'sub-project/package-lock.json'
+    cache-dependency-path: 'sub-project/package-lock.json'
 ```
 ```yml
 steps:
@@ -44,5 +44,5 @@ steps:
   with:
     node-version: 14
     cache: yarn
-    package-lock-file: 'sub-project/**/yarn.lock'
+    cache-dependency-path: 'sub-project/**/yarn.lock'
 ```
--- a/docs/adrs/0001-support-caching-deps-for-monorepos.md
+++ b/docs/adrs/0001-support-caching-deps-for-monorepos.md
@@ -23,8 +23,8 @@ The second option looks more generic because it allows to:
 
 ## Decision
 
-Add `deps-lock-file` input that will accept path (relative to repository root) to dependencies lock file.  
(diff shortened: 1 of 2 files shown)


docs/adrs/0001-support-caching-deps-for-monorepos.md
```diff
@@ -0,0 +1,48 @@
...
+Date: 2021-07-13
+
+Status: Proposed
+
+## Context
+Currently, `actions/setup-node` supports caching dependencies for Npm and Yarn package managers.  
+For the first iteration, we have decided to not support cases where `package-lock.json` / `yarn.lock` are located outside of repository root.  
+Current implementation searches the following file patterns in the repository root: `package-lock.json`, `yarn.lock` (in order of resolving priorities)
+
+Obviosly, it made build-in caching unusable for mono-repos and repos with complex structure.
```
AlenaSviridenko, line 11:
```suggestion
Obviously, it made build-in caching unusable for mono-repos and repos with complex structure.  
```

docs/adrs/0001-support-caching-deps-for-monorepos.md
```diff
@@ -0,0 +1,48 @@
...
+
+The second option looks more generic because it allows to:
+- specify multiple dependencies files using file patterns like `**/package-lock.json` ([one of recommended approaches in actions/cache](https://github.com/actions/cache/blob/main/examples.md#macos-and-ubuntu))
+- specify custom dependencies files like `src/npm-shrinkwrap.json`
+- change default resolving priority if both `yarn.lock` and `package-lock.json` exist in repository
+
+## Decision
+
+Add `deps-lock-file` input that will accept path (relative to repository root) to dependencies lock file.  
+If path provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/packages.lock.json') }}` YAML construction does.
```
AlenaSviridenko, line 27:
```suggestion
If provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/packages.lock.json') }}` YAML construction does.  
```

docs/adrs/0001-support-caching-deps-for-monorepos.md
```diff
@@ -0,0 +1,48 @@
...
+The second option looks more generic because it allows to:
+- specify multiple dependencies files using file patterns like `**/package-lock.json` ([one of recommended approaches in actions/cache](https://github.com/actions/cache/blob/main/examples.md#macos-and-ubuntu))
+- specify custom dependencies files like `src/npm-shrinkwrap.json`
+- change default resolving priority if both `yarn.lock` and `package-lock.json` exist in repository
+
+## Decision
+
+Add `deps-lock-file` input that will accept path (relative to repository root) to dependencies lock file.  
+If path provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/packages.lock.json') }}` YAML construction does.  
+The hash of privided matched files will be used as a part of cache key.
```
AlenaSviridenko, line 28:
```suggestion
The hash of provided matched files will be used as a part of cache key.
```

docs/adrs/0001-support-caching-deps-for-monorepos.md
```diff
@@ -0,0 +1,48 @@
...
+
+Yaml examples:
+```yml
+steps:
+- uses: actions/checkout@v2
+- uses: actions/setup-node@v2
+  with:
+    node-version: 14
+    cache: npm
+    package-lock-file: 'sub-project/package-lock.json'
```
ahdbilal, line 38: Should we use a more generic name such as 'cache-path' so that we can use the same for other setup-actions? Also, we probably don't need apostrophes here.

docs/adrs/0001-support-caching-deps-for-monorepos.md
```diff
@@ -0,0 +1,48 @@
...
+For the first iteration, we have decided to not support cases where `package-lock.json` / `yarn.lock` are located outside of repository root.  
+Current implementation searches the following file patterns in the repository root: `package-lock.json`, `yarn.lock` (in order of resolving priorities)
+
+Obviously, it made build-in caching unusable for mono-repos and repos with complex structure.  
+We would like to revisit this decision and add customization for dependencies lock file location.
+
+## Proposal
+We have the following options:
+1. Allow to specify  directory where `package-lock.json` or `yarn.lock` are located
+2. Allow to specify path to the dependencies lock file (including directory path and filename)
```
ahdbilal, line 17: Will option 2 be modeled similar to the 'path' keyword used in actions/cache?

docs/adrs/0001-support-caching-deps-for-monorepos.md
maxim-lobanov: I think `cache-path` can confuse users because it sounds like a path to cache or folder that will be cached. Actually, this variable points to dependencies lock file (file that describes / define dependencies of the project). IMHO, `package-lock-file` or `package-lock-path` is generic enough to be used across other actions too. Another option - something like `deps-lock-path`.

docs/adrs/0001-support-caching-deps-for-monorepos.md
maxim-lobanov: Nope, `path` input in actions/cache defines what folder will be cached. Users of setup-node don't need to care about it. It is done under hood from our side.
This is more about postfix of `key` input, see: 
> key: ${{ runner.os }}-${{ hashFiles('**src/package-lock.json**') }}

User specifies path to file that describes their dependencies and hash of this file will be used as a part of key.

dmitry-shibanov (approved):

AlenaSviridenko (approved):

MaksimZhukov (approved):

docs/adrs/0001-support-caching-deps-for-monorepos.md
konradpabjan: Personally I'm not a fan of `cache-path` for the same reasons.

My vote goes to this input being called `cache-dependency-path`.
 - It's prefixed with `cache` so it's immediately obvious what it's for. It will also make the YAML more readable with 2 `cache-*` input keys right next to each other.
- Since we will be supporting wildcards for multiple paths, it would be confusing to have something with a singular `*-file` in the name be used to cache multiple file**s**. This observation generally leads me to prefer having `*-path` in the name vs `*-file` or `*-files`.
- Generic enough where it can be used in other actions. For node we're dealing with lock files, but to be as generic as possible we should avoid having `lock` in the name since other first party actions like Python have different dependency mechanisms. `dependency` in the input key feels like a good middle ground that will make this easily reusable across all first party actions

maxim-lobanov: I agree with Konrad arguments above.

bryanmacfarlane: +1 agreed on points above

bryanmacfarlane (approved):

docs/adrs/0001-support-caching-deps-for-monorepos.md
```diff
@@ -0,0 +1,48 @@
...
+
+The second option looks more generic because it allows to:
+- specify multiple dependencies files using file patterns like `**/package-lock.json` ([one of recommended approaches in actions/cache](https://github.com/actions/cache/blob/main/examples.md#macos-and-ubuntu))
+- specify custom dependencies files like `src/npm-shrinkwrap.json`
+- change default resolving priority if both `yarn.lock` and `package-lock.json` exist in repository
+
+## Decision
+
+Add `cache-dependency-path` input that will accept path (relative to repository root) to dependencies lock file.  
+If provided path contains wildcards, the action will search all maching files and calculate common hash like `${{ hashFiles('**/package-lock.json') }}` YAML construction does.
```
morki, line 27: How is common hash helping the situation? I think it makes it worse, because if I have monorepo with 10 projects and only one `package-lock.json` file is changed, entire cache is not usable. Or am I missing something?

docs/adrs/0001-support-caching-deps-for-monorepos.md
ahdbilal: I meant, will the input syntax and structure of option 2 be similar to what we currently provide with 'path'. 

```
cache-dependency-path: |
  - xyz/requirements.txt
  - abc/requirements.txt
```

docs/adrs/0001-support-caching-deps-for-monorepos.md
maxim-lobanov: The common hash is only one of possible options. You can also specify path to the single file.
Let's imagine that you have the following repo structure:
```
- service1
--- package.json
--- package-lock.json
--- ...
- service2
--- package-lock.json
--- ...
```

In your first workflow, where you build `service1`, you will invoke setup-node as
```yml
- uses: actions/setup-node@v2
  with:
    cache: npm
    cache-dependency-path: service1/package-lock.json
```
In the second workflow, you will invoke setup-node as
```yml
- uses: actions/setup-node@v2
  with:
    cache: npm
    cache-dependency-path: service2/package-lock.json
```

So every workflow will use own cache and if one `package-lock.json` is changed, only one cache will be regenerated. 

Option with wildcard makes sense when you build multiple services in scope of one workflow, so you will have common cache for a few services. This approach is recommended here: https://github.com/actions/cache/blob/main/examples.md#macos-and-ubuntu

ADR contains a bunch of other examples.
Hope I answered your question.

morki: Yes, thank you very much. I am looking forward to it, looks pretty amazing :)

docs/adrs/0001-support-caching-deps-for-monorepos.md
maxim-lobanov: Hm, I think wildcard will resolve such cases too but I don't have objections to support both options:
```yml
cache-dependency-path: xyz/requirements.txt
```
or
```yml
cache-dependency-path: |
  xyz/requirements.txt
  abc/requirements.txt
```

ahdbilal (approved):

docs/adrs/0001-support-caching-deps-for-monorepos.md
andrewshawcare: Just curious... why name it `cache-dependency-path` and not something like `cache-lock-file-glob`?

It isn't really a path to dependencies is it?

maxim-lobanov: https://github.com/actions/setup-node/pull/299#discussion_r669502996

andrewshawcare: Okay, interesting.

Why make it generic to accommodate python? This is `setup-node`... Is there a plan to make this action handle caching across all languages? If not I would argue that being generic here leads to confusion. It's not obvious from the name that you're looking for the location of my lock files (as opposed to `package.json` files, or `node_modules` folders).

Also, path doesn't really handle what is being done here... you're using a glob. This can cover multiple files _and multiple paths_.

michielswaanen:
Looks good! Any update on this? Really eager to see this feature released

AlexITC:
Any plans to merge this soon?

docs/adrs/0001-support-caching-deps-for-monorepos.md
MaksimZhukov: Thank you for sharing your opinion. We prefer to use `cache-dependency-path` as the input name for the reasons listed in [this comment](https://github.com/actions/setup-node/pull/299#discussion_r669502996).

PR #305, opened 2021-08-02
Support caching for mono repos and repositories with complex structure

MaksimZhukov:
In scope of this PR, we are adding the `cache-dependency-path` input that accepts path (relative to repository root) to dependencies file.  Path can contain wildcards patterns and a list of file paths. For more information, please see the related [ADR](https://github.com/actions/setup-node/pull/299).


README.md
```diff
@@ -90,7 +90,7 @@ steps:
 - run: pnpm test
 ```
 
-> At the moment, only `lock` files in the project root are supported.
+For more examlpes of caching, please see the [Advanced usage](docs/advanced-usage.md#caching-packages-dependencies) guide.
```
maxim-lobanov, line 93:
```suggestion
For more examples of caching, please see the [Advanced usage](docs/advanced-usage.md#caching-packages-dependencies) guide.
```

maxim-lobanov, line 93: I think we should remove this note because currently it looks like it is related to only pnpm

README.md
```diff
@@ -41,7 +41,7 @@ nvm lts syntax: `lts/erbium`, `lts/fermium`, `lts/*`
 
 ### Caching packages dependencies
 
-The action has a built-in functionality for caching and restoring npm/yarn dependencies. Supported package managers are `npm`, `yarn`, `pnpm`. The `cache` input is optional, and caching is turned off by default.
+The action has a built-in functionality for caching and restoring npm/yarn dependencies. Supported package managers are `npm`, `yarn`, `pnpm`. The `cache` input is optional, and caching is turned off by default. By default, the action searches for the dependency file in the project root and uses its hash as a part of cache key. Use `cache-dependency-path` to specify custom dependency file path. The field accepts wildcards or an array of files to be cached.
```
maxim-lobanov, line 44:
```suggestion
The action has a built-in functionality for caching and restoring dependencies.  It uses [actions/cache](https://github.com/actions/cache) under hood for caching dependencies but requires less configuration settings.
Supported package managers are `npm`, `yarn`, `pnpm`. The `cache` input is optional, and caching is turned off by default.
As recommended by [actions/cache](https://github.com/actions/cache/blob/main/examples.md#node---npm), action caches global cache on machine instead of `node_modules` so cache can be reused between different Node.js versions.  
By default, the action searches for the dependency file (`package-lock.json` or `yarn.lock`) in the repository root and uses its hash as a part of cache key. Use `cache-dependency-path` for cases where multiple dependency files are used for build or they are located in  subdirectories. See examples of `cache-dependency-path` usage in the [Advanced usage](docs/advanced-usage.md#caching-packages-dependencies) guide.
```

AlenaSviridenko, line 44: In general, I like the Maxim's suggestion, but I'd like to propose some changes

The action has a built-in functionality for caching and restoring dependencies.  It uses [actions/cache](https://github.com/actions/cache) under hood for caching dependencies but requires less configuration settings.
Supported package managers are `npm`, `yarn`, `pnpm`. The `cache` input is optional, and caching is turned off by default.  
The action defaults to search for the dependency file (`package-lock.json` or `yarn.lock`) in the repository root, and uses its hash as a part of the cache key. Use `cache-dependency-path` for cases when multiple dependency files are used, or they are located in different subdirectories. See the examples of `cache-dependency-path` usage in the [Advanced usage](docs/advanced-usage.md#caching-packages-dependencies) guide.  
The action follows [actions/cache](https://github.com/actions/cache/blob/main/examples.md#node---npm) guidelines, and caches global cache on the machine instead of `node_modules`, so cache can be reused between different Node.js versions.

AlenaSviridenko (approved):

README.md
MaksimZhukov: Thank you!

README.md
MaksimZhukov: Thanks!

action.yml
```diff
@@ -21,6 +21,8 @@ inputs:
     default: ${{ github.token }}
   cache:
     description: 'Used to specify a package manager for caching in the default directory. Supported values: npm, yarn, pnpm'
+  cache-dependency-path:
+    description: 'Used to specify path to a dependencies file: package-lock.json, yarn.lock, etc. Supports wildcards or an array of file names.'
```
konradpabjan, line 25:
```suggestion
    description: 'Used to specify the path to a dependency file: package-lock.json, yarn.lock, etc. Supports wildcards or a list of file names for caching multiple dependencies.'
```

`List` makes more sense vs `array` since you can't do something like

```
    cache-dependency-path: |
      [server/app/package-lock.json
      frontend/app/package-lock.json]
```

src/cache-restore.ts
```diff
@@ -22,9 +25,17 @@ export const restoreCache = async (packageManager: string) => {
     packageManagerInfo,
     packageManager
   );
-  const lockFilePath = findLockFile(packageManagerInfo);
+  const lockFilePath = cacheDependencyPath
+    ? cacheDependencyPath
```
konradpabjan, line 29: Question.

It looks like the raw `cacheDependencyPath` path input is used as the input to `glob.hashFiles`.

Lets imagine I have 2 different inputs that lead to the same dependency files.

In the first I have
```
    cache-dependency-path: |
      server/app/package-lock.json
      frontend/app/package-lock.json
```

While in the second i use
```
    cache-dependency-path: **/*package-lock.json
```

Would the `fileHash` generated by both inputs be the same?

maxim-lobanov (approved):

vsafonkin (approved):

src/cache-restore.ts
MaksimZhukov: If the repository only contains these two `package-lock.json` files, the `fileHash` will be the same for both cases.

konradpabjan (approved):

Thompson1985 (approved):

PR #308, opened 2021-08-03
Temporarily disabled cache usage for v2

AlenaSviridenko:
Temporarily disabling usage of pre-cached Node.js versions


MaksimZhukov (approved):

HonkingGoose:
> Temporarily disabling usage of pre-cached Node.js versions

I just got a Renovate bot PR that wants to update `action/setup-node` from `v2.3.0` to `v2.3.1`, and I'm really confused on _why_ I as a end-user should update to the newer version.

Can somebody explain in a general sense what's going wrong right now, and how the temporary fix helps with this problem?

I've read the PR description, and checked the latest release notes, but they don't have a explanation at all.

PR #323, opened 2021-08-26
feat(cache): change key to match actions/cache documentation

mmorel-35:
This PR change the cache key construction to fit [those examples](https://github.com/actions/cache/blob/main/examples.md#node---npm). It also add the use of restoreKeys and a more verbose log when retrieving fails.

### DIFF
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -135,10 +135,10 @@ describe('cache-restore', () => {
         await restoreCache(packageManager);
         expect(hashFilesSpy).toHaveBeenCalled();
         expect(infoSpy).toHaveBeenCalledWith(
-          `Cache restored from key: node-cache-${platform}-${packageManager}-${fileHash}`
+          `Cache restored from key: ${platform}-setup-node-${packageManager}-${fileHash}`
         );
         expect(infoSpy).not.toHaveBeenCalledWith(
-          `${packageManager} cache is not found`
+          `Cache not found for input keys: ${platform}-setup-node-${packageManager}-${fileHash}, ${platform}-setup-node-${packageManager}-, ${platform}-setup-node-`
         );
       }
     );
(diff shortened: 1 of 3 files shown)


oxc:
Does it make sense to fallback on a shared prefix without the package manager? They probably even use different cache directories.

ThisIsMissEm:
> Does it make sense to fallback on a shared prefix without the package manager? They probably even use different cache directories.

I would agree, falling back to a prefix without package manager probably isn't useful — they definitely use separate cache paths.

belgattitude:
Hey just saw a link to this PR. I'd like to mention few discoveries I did recently

- Yarn 3+ cache is automatically pruned when deps changes (it does not grow over time on the github cache).
- Yarn 3+ cache is OS/arch independent (no need to prefix with OS)
- Pnpm 7 is not pruned automatically,. There's a specific command https://pnpm.io/cli/store#prune. It's sometimes buggy.

You can see two examples based on `@action/cache`:

- yarn: https://gist.github.com/belgattitude/042f9caf10d029badbde6cf9d43e400a
- pnpm: https://gist.github.com/belgattitude/838b2eba30c324f1f0033a797bab2e31

It might be helpful.

oxc:
> * Yarn 3+ cache is automatically pruned when deps changes (it does not grow over time on the github cache).

I believe this is only true if you're not using the [global cache](https://yarnpkg.com/configuration/yarnrc#enableGlobalCache)  in your project. From `$(yarn config get cacheFolder)` alone there is no way to distinguish the two.
See also https://yarnpkg.com/features/offline-cache#sharing-the-cache

belgattitude:
True, if you look at the yarn gist you'll see that on ci the global cache is disabled YARN_ENABLE_GLOBAL_CACHE=false through an env. That makes it work independently of config.

oxc:
> True, if you look at the yarn gist you'll see that on ci the global cache is disabled YARN_ENABLE_GLOBAL_CACHE=false through an env. That makes it work independently of config.

I see. For the record, I would hate it if the node action overrode my yarn config 😄

Thompson1985 (approved):

PR #326, opened 2021-09-03
Use CI friendly commands in documentation

jonkoops:
Replaces `npm install` with [`npm ci`](https://docs.npmjs.com/cli/v7/commands/npm-ci) command and adds the `--frozen-lockfile` flag for [Yarn](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-frozen-lockfile) and [PNPM](https://pnpm.io/cli/install#--frozen-lockfile) installation commands. This makes installation of packages faster (for NPM) and provides guarantees about lockfile correctness at install time.

Previous efforts for this are documented in https://github.com/actions/setup-node/pull/103.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -1,4 +1,38 @@
-# Advanced usage
+## Working with lockfiles
+
+All supported package managers recommend that you **always** commit the lockfile, although implementations vary doing so generally provides the following benefits:
+
+- Enables faster installation for CI and production environments, due to being able to skip package resolution.
+- Describes a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
+- Provides a facility for users to "time-travel" to previous states of `node_modules` without having to commit the directory itself.
+- Facilitates greater visibility of tree changes through readable source control diffs.
+
+In order to get the most out of using your lockfile on continuous integration follow the conventions outlined below for your respective package manager.
+
+### NPM
+
+Ensure that `package-lock.json` is always committed, use `npm ci` instead of `npm install` when installing packages.
+
+**See also:**
+- [Documentation of `package-lock.json`](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json)
+- [Documentation of `npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci)
+
+### Yarn
+
+Ensure that `yarn.lock` is always committed, pass `--frozen-lockfile` to `yarn install` when installing packages.
+
+**See also:**
+- [Documentation of `yarn.lock`](https://classic.yarnpkg.com/en/docs/yarn-lock)
+- [Documentation of `--frozen-lockfile` option](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile)
+- [QA - Should lockfiles be committed to the repoistory?](https://yarnpkg.com/getting-started/qa/#should-lockfiles-be-committed-to-the-repository)
+
+### PNPM
+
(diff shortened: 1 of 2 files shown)


privatenumber:
@dmitry-shibanov Anything blocking this from getting merged?

pi9min:
+1

docs/advanced-usage.md
```diff
@@ -86,7 +86,7 @@ steps:
     node-version: '14'
     cache: 'npm'
     cache-dependency-path: '**/package-lock.json'
-- run: npm install
+- run: npm ci
```
hubgit, line 89: Does it make sense to use `npm ci` when the dependencies are being restored from a cache?

jonkoops: I believe the cache option only caches the global cache, not the `node_module` directory. So it still makes sense.

hubgit: It does look like it's running `npm config get cache` to get the path to cache, so that does make sense, thanks 👍

jonkoops:
@dmitry-shibanov @brcrista @marko-zivic-93 is there any interest in getting this merged?

dmitry-shibanov:
Hello @jonkoops. Thank you for your pull request. I think we can adjust these changes when caching is enabled, because it relies on hash from dependency files. For other cases I think we should leave it as `npm install` because some repositories do not have package-lock.json.

jonkoops:
> For other cases I think we should leave it as `npm install` because some repositories do not have package-lock.json.

I think differently about this matter, checking in lockfiles is considered best practice for all package managers. Instead of appeasing users that might not have checked this file in we should encourage them to follow these practices.

For example, the [NPM documentation](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json) says the following about the `package-lock.json` file:

> This file is **intended to be committed into source repositories**, and serves various purposes:
> - Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> - Provide a facility for users to "time-travel" to previous states of `node_modules` without having to commit the directory itself.
> - Facilitate greater visibility of tree changes through readable source control diffs.
> - Optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.
>  - As of npm v7, lockfiles include enough information to gain a complete picture of the package tree, reducing the need to read `package.json` files, and allowing for significant performance improvements.

All of these points apply directly to Github and Github Actions and I believe that these best practices should be reflected here as well.

I propose that all commands are replaced with their appropriate CI friendly versions as per the original purpose of this PR and that the documentation is amended to have a dedicated section to lockfiles.

jonkoops:
I've added some documentation about lockfiles as mentioned above.

privatenumber:
For an official action like this, I think it's important to use examples that adopt best practices.

There are many beginners that read these examples that are not aware of the trade-offs between `npm install` and `npm ci`.

The problem with using `npm install` or not using `package-lock.json` is that there's a chance their CI will unexpectedly fail because of inconsistent dependencies.

For example, a new dependency update that satisfies the `package.json` semver range can be installed on the CI before the developer tests it locally and fail (or even worse, fail silently and trigger a release).

[`npm ci` was designed for Continuous Integration](https://docs.npmjs.com/cli/v8/commands/npm-ci#:~:text=it%27s%20meant%20to%20be%20used%20in%20automated%20environments%20such%20as%20test%20platforms%2C%20continuous%20integration%2C%20and%20deployment) to prevent these problems.

dmitry-shibanov (approved):

dmitry-shibanov:
Thank you @jonkoops for improving the documentation and your work.

jonkoops:
Thanks for taking this into consideration and merging it!

PR #368, opened 2021-11-29
Print node, npm and yarn versions after installation

havenchyk:
**Description:**
Print node, npm and yarn version as it was before https://github.com/actions/setup-node/pull/147

**Related issue:**
#90

**Check list:**


src/main.ts
```diff
@@ -105,3 +108,17 @@ function resolveVersionInput(): string {
...
+async function printEnvDetails() {
+  // Output version of node and npm that are being used
+  await exec.exec('node', ['--version']);
+
+  // Older versions of Node don't include npm, so don't let this call fail
+  await exec.exec('npm', ['--version'], {
+    ignoreReturnCode: true
+  });
+
+  await exec.exec('yarn', ['--version'], {
```
havenchyk, line 121: probably yarn is not need as it doesn't go with node, is that correct?

nickmccurdy: Both GitHub's runners and corepack (if enabled) will automatically install Yarn

IvanZosimov:
Hi, @havenchyk 👋 ! Thank you for your contribution! I'd suggest using [getExecOutput()](https://github.com/actions/toolkit/blob/main/packages/exec/src/exec.ts#L44) function instead of exec() to be able to format the information in the log and also to handle such situations when we want to notify user that there is no such util in the PATH (for instance the customer changed the PATH on the self-hosted runner and erased npm). Probably it'd be interesting to play with putting version information into the [foldable group](https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L301-L310).

IvanZosimov:
📟 Just a gentle ping, @havenchyk

havenchyk:
hi guys, definitely I'll take a look: resolve conflicts today-tomorrow, thanks for reminders

havenchyk:
@dmitry-shibanov @IvanZosimov rebased, sorry for the delay

havenchyk:
oh, just noticed a comment about refactoring, let me address it

havenchyk:
@IvanZosimov can you please help me with the testing of changes? I see actions require some approval to run

dmitry-shibanov:
Hello @havenchyk. Could you please run `npm run build` and `npm run format` commands. Besides, I think we need to move yarn and npm to the try-cacth block too because it can cause issues like [this](https://github.com/actions/setup-node/issues/538) for docker images when the tool does not exist or it is not supported.

havenchyk:
@dmitry-shibanov done. I think we need to run some action to verify what will the code do. Before I fix the tests, I know they will fail

marko-zivic-93 (approved):

havenchyk:
Guys, I need help with testing :) which job should I check to understand that the version is printed correctly?

src/main.ts
```diff
@@ -108,3 +98,47 @@ function resolveVersionInput(): string {
 
   return version;
 }
+
+async function printEnvDetailsAndSetOutput() {
+  core.startGroup('Environment details');
+  // Output version of node is being used
+  try {
```
dmitry-shibanov, line 105: Could please create a separate function getToolVersion:
```
async function getToolVersion(tool: string, options: string[]) {
  try {
    const {stdout, stderr, exitCode} = await exec.getExecOutput(tool, options, {
      ignoreReturnCode: true,
      silent: true
    });

    if (exitCode) {
      core.info(`[warning]${stderr}`);
      return '';
    }

    return stdout;
  } catch (err) {
    return '';
  }
}
```

__tests__/installer.test.ts
```diff
@@ -249,6 +249,21 @@ describe('setup-node', () => {
 
     let expPath = path.join(toolPath, 'bin');
 
+    expect(getExecOutputSpy).toHaveBeenCalledWith(
```
havenchyk, line 252: @dmitry-shibanov don't you think such check is better?

src/main.ts
```diff
@@ -108,3 +98,35 @@ function resolveVersionInput(): string {
...
+export async function printEnvDetailsAndSetOutput() {
+  core.startGroup('Environment details');
+
+  const promises = ['node', 'npm', 'yarn'].map(async tool => {
+    const output = await getToolVersion(tool, ['--version']);
+
+    core.setOutput(`${tool}-version`, output);
+  });
+
+  await Promise.all(promises);
```
havenchyk, line 111: I didn't check it 😄

__tests__/installer.test.ts
```diff
@@ -41,6 +41,7 @@ describe('setup-node', () => {
   let parseNodeVersionSpy: jest.SpyInstance;
   let isCacheActionAvailable: jest.SpyInstance;
   let getExecOutputSpy: jest.SpyInstance;
+  let execExecSpy: jest.SpyInstance;
```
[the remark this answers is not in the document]
havenchyk: removed

src/main.ts
havenchyk: done (a bit different)

src/main.ts
```diff
@@ -108,3 +98,35 @@ function resolveVersionInput(): string {
...
   return version;
 }
+
+export async function printEnvDetailsAndSetOutput() {
+  core.startGroup('Environment details');
+
+  const promises = ['node', 'npm', 'yarn'].map(async tool => {
+    const output = await getToolVersion(tool, ['--version']);
+
+    core.setOutput(`${tool}-version`, output);
```
dmitry-shibanov, line 108: node-version set on previous steps could you please remove it from there. Besides, could you please logs for versions ?

havenchyk: I don't fully understand your comment. Please check those changes https://github.com/actions/setup-node/pull/368/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dL43-L53

dmitry-shibanov: Sorry, I're right. They were removed. Could you please add core.info to print versions ?

havenchyk: Added. Did you mean this? Should I keep that `core.setOutput(`${tool}-version`, output);`?

dmitry-shibanov: According to the issue it's only about logging. I think for now we can remove output If we get separate issue about output we'll add it. Thank you.

havenchyk: the previous version had `node-version` in the output, should we keep it?

dmitry-shibanov: For node-version we should keep it.

havenchyk: Done

havenchyk:
@dmitry-shibanov please take a look once again

marko-zivic-93 (approved):

vsafonkin (approved):

__tests__/installer.test.ts
Jajajjshsdhdhgekskaooqoq: Thank you

src/main.ts
Jajajjshsdhdhgekskaooqoq: Thank yo

dmitry-shibanov:
Hello @havenchyk. Thank you for contribution. Sorry for the late response.

PR #373, opened 2021-12-02
Add support for asdf format as Node.js version file

ganta:
**Description:**
I'm happy to be supported the Node.js version file in #338.
However, we're migrating Node.js version management to asdf.

The version specification file used in asdf has the following format, which is not supported by the current `node-version-file` input:

```
nodejs 14.18.2
```
```
ruby 2.7.4
nodejs 14.18.2
python 3.8.12
```

So, I implemented `node-version-file` to support the format asdf's `.tool-versions` file.

**Related issue:**
#222

**Check list:**
- [x] Mark if documentation changes are required.
- [x] Mark if tests were added or updated to cover the changes.

PR #445, opened 2022-03-15
Allow using .tool-versions for node-version-file

deiga:
**Description:**
Add parsing of `.tool-versions` file format.
This enables developers which use `asdf` to use their version file to
steer which version of node to install.
This change should be fully backwards compatible as it only does things
if a certain string is found in the `node-version-file`

**Check list:**


Johnstone456 (approved):

brcrista:
Hey @deiga, thanks for the PR.  This is my first time hearing about `asdf`, but I think the approach makes sense.  If I'm understanding correctly, you don't actually have to install `asdf` on the build machine; this is just adding support for its syntax.  Is that right?

Since this is a new feature, if you're able to put together a minimal workflow that uses `deiga/setup-node`, that would go a long way towards helping us see how it works.

runlevel5:
Thanks for the PR, we are also using `asdf` for our projects (due to mix-tech nature). Love to see this PR going in

IvanZosimov:
@deiga, 👋 Just a gentle ping 💌

dmitry-shibanov:
Hello @deiga. Thank you for your pull request. For now I'm going to close the pull request because the logic was implemented in this [pull request](https://github.com/actions/setup-node/pull/373). If you have any concerns feel free to ping us.

PR #470, opened 2022-04-19
extend documentation on `scope` parameter

tyrann0us:
**Description:**
In [`writeRegistryToFile()`](https://github.com/actions/setup-node/blob/6e4a7921f09320d9891f6ccda63804def2ff9da2/lib/authutil.js#L23-L52) (added in https://github.com/actions/setup-node/pull/21/files#diff-a3c83546a6fde2740e1dd05c8267c74af1c8a7fe3add285eea3669b3760ac6f3R25-R27), `scope` falls back to the repository owner if the `npm.pkg.github.com` registry (GitHub Packages) is used.
Since in most cases, this means that the `scope` parameter is not needed, but the logic is not documented anywhere, this PR adds the appropriate note to the parameter description.

**Related issue:**
None.

**Check list:**

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,35 +1,35 @@
 name: 'Setup Node.js environment'
-description: 'Setup a Node.js environment by adding problem matchers and optionally downloading and adding it to the PATH'
+description: 'Setup a Node.js environment by adding problem matchers and optionally downloading and adding it to the PATH.'
 author: 'GitHub'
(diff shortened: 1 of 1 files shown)


brcrista (approved):
thanks!

brcrista:
@dmitry-shibanov FYI

Issue #480, opened 2022-05-02
Enable Corepack


shellscape:
**Description:**

Related to this issue and comment https://github.com/actions/setup-node/issues/182#issuecomment-1086296667, I'd like to see support for enabling [corepack](https://nodejs.org/api/corepack.html) without having to have a separate `run` step to do so.

**Justification:**

This would reduce overall runtime of actions, as no separate steps are required to enable corepack nor to install package managers, which is a net benefit to users in both regards. 

**Are you willing to submit a PR?**

Surely

e-korolevskii:
Hi @shellscape,
We will investigate the possibility of adding this feature.

e-korolevskii:
Hi @shellscape,
Unfortunately, corepack is experimental now, so we would not implement that at least while its not become stable.
That is why I will close this issue now. If you have any questions or concerns, please feel free to leave a comment on this thread or open another issue.

styfle:
A PR was already started in #482 

Also, we should get clarity around the path from experimental to stable https://github.com/nodejs/corepack/issues/104

dmitry-shibanov:
Hello @shellscape. I think the main reason to not enable corepack by default because some customers can use setup-node with cache enable in conjunction with pnpm/action-setup before setup-node. When corepack is enabled it will override preinstalled pnpm (from pnpm/action-setup).
For example users can specify use the latest version of pnpm (7th), but corepack can override it to 6th. This also applies to yarn.

Ethan-Arrowood:
corepack will not override it to 6 unless that is what is set for `packageManager`

dmitry-shibanov:
Hello @Ethan-Arrowood. Thank you for your response. I've tried to use this snippet and it overrides pnpm version.
```
steps:
  - name: Checkout repository
    uses: actions/checkout@v3
  - uses: pnpm/action-setup@v2
    with:
      version: 7
  - run: pnpm --version
  - uses: actions/setup-node@v3
    with:
      node-version: 16.9.0
  - run: |
      corepack enable
      pnpm --version
```

Ethan-Arrowood:
Is `packageManager` in your package.json set to `pnpm@7.0.0`?

dmitry-shibanov:
> Is `packageManager` in your package.json set to `pnpm@7.0.0`?

No it does not. With packageManager it works as expected. But not all users have packageManager field, that is why it can break existing behaviour some users if corepack is enabled by default with caching.

viceice:
[...]

PR #482, opened 2022-05-02
Add automatic corepack support for node v14 +

Ethan-Arrowood:
**Description:**

Executes `corepack enable` when the detected node version is 14+. This will enable support for the `packageManager` field for `yarn` and `pnpm`

**Related issue:**
https://github.com/actions/setup-node/issues/480

**Check list:**

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,8 +4,9 @@ import fs from 'fs';
 import * as auth from './authutil';
 import * as path from 'path';
 import {restoreCache} from './cache-restore';
-import {isGhes, isCacheFeatureAvailable} from './cache-utils';
+import {isGhes, isCacheFeatureAvailable, getCommandOutput} from './cache-utils';
 import os = require('os');
+import semver from 'semver';
 
 export async function run() {
   try {
@@ -46,6 +47,9 @@ export async function run() {
     }
 
     if (cache && isCacheFeatureAvailable()) {
+      if (semver.gte(version, '14')) {
+        await getCommandOutput('corepack enable');
+      }
       const cacheDependencyPath = core.getInput('cache-dependency-path');
       await restoreCache(cache, cacheDependencyPath);
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@actions/http-client": "^1.0.11",
         "@actions/io": "^1.0.2",
         "@actions/tool-cache": "^1.5.4",
-        "semver": "^6.1.1"
+        "semver": "^6.3.0"
       },
       "devDependencies": {
         "@types/jest": "^27.0.2",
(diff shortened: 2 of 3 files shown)


src/main.ts
```diff
@@ -46,6 +47,9 @@ export async function run() {
     }
 
     if (cache && isCacheFeatureAvailable()) {
+      if (semver.gte(version, '14')) {
```
styfle, line 50: This needs to be `>=14.19.0`

https://nodejs.org/api/corepack.html

src/main.ts
```diff
@@ -46,6 +47,9 @@ export async function run() {
...
+        await getCommandOutput('corepack enable');
```
styfle, line 51: Maybe this should try/catch?

Ethan-Arrowood:
Struggling to get a test to pass. Insight appreciated

Ethan-Arrowood:
Probably wrong on the test, but since it is asserting the desired corepack output I'm marking this PR ready for review. Hopefully a maintainer more familiar with the test suite can provide some input on how to improve the tests.

shellscape:
Thanks for opening this PR. I support the idea.

styfle (commented):
We also need to consider caching the result of corepack so that it doesn't need to install the package manger each time.

Ethan-Arrowood:
Added a check for the `packageManager` field. How do I fake a `package.json` file for the test? 

I almost want to try writing an integration test for this, but would appreciate some input from the maintainers

imanuel112 (approved):

imanuel112 (approved):

nickmccurdy (changes requested):
I'm in favor of using Corepack, but have some concerns about the implementation:
1. ~This doesn't seem to support the `@version` syntax, which is automatically used in Yarn 3.1, and documented as experimental in Node. We'd have to at least allow for this syntax to set the package manager properly, though it would be nice to also parse the version and use it by default.~
2. GitHub runners already have Yarn installed. You're technically supposed to uninstall Yarn before enabling Corepack, as the global Yarn binary would conflict with Corepack's Yarn shim. We could try to fix this by either uninstalling Yarn before enabling Corepack, or removing Yarn from GitHub's runner (which would be a breaking change but would standardize behavior in the long run).

styfle:
@nickmccurdy 

1. Can you explain more about this version syntax? I’m pretty sure “corepack enable will work with nearly all versions of Yarn.
2. I think it would be hard to remove yarn from the runner so perhaps it should be uninstalled. What’s the best way to do that? npm uninstall -g yarn?

nickmccurdy:
> 1. Can you explain more about this version syntax? I’m pretty sure “corepack enable will work with nearly all versions of Yarn.

Oh sorry, I see what you mean. It's not actually parsing the `packageManager`, but rather passing it to corepack.

> 2. I think it would be hard to remove yarn from the runner so perhaps it should be uninstalled. What’s the best way to do that? npm uninstall -g yarn?

GitHub Actions seems to run `npm install -g yarn ...`, so yea, `npm uninstall -g yarn` should work.

jtbandes:
@Ethan-Arrowood Are you planning to update this PR based on the feedback that has been provided? I know a lot of people are still interested in this feature. I see that CI is also failing currently.

styfle:
Looking at this PR again, it could be considered semver major because it will suddenly opt into corepack when the package.json is detected. 

I'm starting to think that PR #651 is better solution because it could be considered semver minor due to the manual opt in.

That could prove valuable if anything changes in the future, since corepack is still technically experimental.

PR #496, opened 2022-05-19
Get latest version from cache if exists

panticmilos:
**Description:**
Small improvement for recent PR related to the latest syntax. There is the possibility that the latest version is actually cached (v18 in the future for example), so it's not needed to download it each time from dist.

**Related issue:**

**Check list:**
- [ x ] Mark if tests were added or updated to cover the changes.

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -7,6 +7,7 @@ import * as tc from '@actions/tool-cache';
 import * as path from 'path';
 import * as semver from 'semver';
 import fs = require('fs');
+import * as installer from './installer';
 
 //
 // Node versions interface
@@ -66,8 +67,8 @@ export async function getNode(
     }
   }
 
-  if(['current', 'latest', 'node'].includes(versionSpec)) {
-    versionSpec = await queryDistForMatch(versionSpec, arch); 
+  if (isLatestSyntax(versionSpec)) {
+    versionSpec = await queryDistForMatch(versionSpec, arch);
     core.info(`getting latest node version...`);
   }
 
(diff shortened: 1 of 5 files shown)


dmitry-shibanov (approved):

src/installer.ts
```diff
@@ -7,6 +7,7 @@ import * as tc from '@actions/tool-cache';
 import * as path from 'path';
 import * as semver from 'semver';
 import fs = require('fs');
+import * as installer from './installer';
```
vsafonkin, line 10: I'm a little confused this line: why we need to import `installer` to the `installer` file?

panticmilos: left by mistake, removed

src/installer.ts
```diff
@@ -370,14 +382,14 @@ async function queryDistForMatch(
       throw new Error(`Unexpected OS '${osPlat}'`);
   }
 
+  if (!nodeVersions) {
+    core.debug('No dist manifest cached');
+    nodeVersions = await getVersionsFromDist();
+  }
+
```
vsafonkin, line 389: I believe it will be easier to invoke `getVersionsFromDist()` once without any conditions in `queryDistForMatch` function. In this case we can remove `getVersionsFromDist()` from this code:
```javascript
if (isLatestSyntax(versionSpec)) {
    nodeVersions = await getVersionsFromDist();
    versionSpec = await queryDistForMatch(versionSpec, arch, nodeVersions);
    core.info(`getting latest node version...`);
  }
```
In this case we no need to get and pass `nodeVersions` array to `queryDistForMatch` function because we use this array only for this condition.

I believe it's more clear logic.

dmitry-shibanov: @vsafonkin In this case we can have a case to call dist api twice, that is why we grab all versions at the beginning and pass it to further functions.

marko-zivic-93 (approved):

PR #502, opened 2022-05-30
Fix to run tests on non-x64 architecture

ganta:
**Description:**
The test fails when run on non-x86 architectures such as Apple Silicon.

The return value of `os.arch()` can be replaced using Jest's `spyOn()` so that tests can be run on non-x86 architectures, but this does not work as expected in certain cases.
https://github.com/actions/setup-node/blob/ed1a46e9f278c17e07705c4cc19a1d9a3fca12ac/__tests__/installer.test.ts#L54-L55

In this case, `os.arch()` is called with the default parameter, so it cannot be replaced.
https://github.com/actions/toolkit/blob/8263c4d15d3a8616851e9c7762ce46fbb4f8c552/packages/tool-cache/src/tool-cache.ts#L638

Therefore, I have changed it to pass values instead of relying on default values.

**Related issue:**

N/A

**Check list:**


vsafonkin (approved):

ganta:
I have resolved the conflict. Could you please reconfirm?

marko-zivic-93 (approved):

Issue #516, opened 2022-06-13
Error: getCacheEntry failed: Cache service responded with 503


mbaum0:
```
2022-06-13T06:46:29.7622187Z ##[group]Run actions/setup-node@v2
2022-06-13T06:46:29.7622425Z with:
2022-06-13T06:46:29.7622601Z   node-version: 16.13
2022-06-13T06:46:29.7622858Z   cache: npm
2022-06-13T06:46:29.7623136Z   cache-dependency-path: ./slack-adapter/package-lock.json
2022-06-13T06:46:29.7623405Z   always-auth: false
2022-06-13T06:46:29.7623606Z   check-latest: false
2022-06-13T06:46:29.7623978Z   token: ***
2022-06-13T06:46:29.7624149Z ##[endgroup]
2022-06-13T06:46:29.9263615Z Attempting to download 16.13...
2022-06-13T06:46:30.1968736Z Acquiring 16.13.2 - x64 from https://github.com/actions/node-versions/releases/download/16.13.2-1681281899/node-16.13.2-linux-x64.tar.gz
2022-06-13T06:46:30.5470929Z Extracting ...
2022-06-13T06:46:30.5583930Z [command]/usr/bin/tar xz --strip 1 --warning=no-unknown-keyword -C /home/runner/work/_temp/1c9fdfcb-b4e4-45cc-b8c0-879d4d188913 -f /home/runner/work/_temp/288baaa0-6ffa-45e6-b176-3504aa37feb4
2022-06-13T06:46:31.3994436Z Adding to the cache ...
2022-06-13T06:46:32.2973296Z Done
2022-06-13T06:46:32.2987622Z [command]/opt/hostedtoolcache/node/16.13.2/x64/bin/npm config get cache
2022-06-13T06:46:32.5559356Z /home/runner/.npm
2022-06-13T06:46:42.7419366Z ##[error]getCacheEntry failed: Cache service responded with 503
```

**Description:**
setup-node fails with cacheing error

**Action version:**
@actions/setup-nodev2

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**

node: 16.13.2

**Repro steps:**  
Workflow file:
```
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2
        with:
          node-version: 16.13
          cache: "npm"
[...]

PR #517, opened 2022-06-13
Fix: Show warning instead of failing the run

cino:
**Description:**
As there are currently issues with the caching service which is breaking workflows using the caching I suggest we change the workflow to only *fail* when there are configuration errors. When anything else goes wrong in the workflow, like a 503 error there should be a warning but let the workflow run continue. 

Having caching available or not should not be breaking the workflow, instead, just continue as if there is no cache. 

**Related issue:**
https://github.com/actions/setup-node/issues/516

**Check list:**

### DIFF
--- a/__tests__/cache-utils.test.ts
+++ b/__tests__/cache-utils.test.ts
@@ -46,7 +46,8 @@ describe('cache-utils', () => {
     isFeatureAvailable.mockImplementation(() => false);
     process.env['GITHUB_SERVER_URL'] = 'https://www.test.com';
 
-    expect(() => isCacheFeatureAvailable()).toThrowError(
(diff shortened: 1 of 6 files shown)


cino:
`Check dist` action was failing due to me not committing the dist files, however, I think someone should update the [contributors.md](https://github.com/actions/setup-node/blob/main/docs/contributors.md)

IvanZosimov:
Hi, @cino 👋 ! Thank you for your contribution a lot! Unfortunately, we decided to implement such error handling by ourselves on another let's say deeper level. We really appreciate your effort and willingness to help! ❤️

By now I'm going to close this PR, if you have any questions or suggestions feel free to ping us.

Issue #541, opened 2022-07-12
bug: wrong node version installed when cache is corrupt


ben-styling:
**Description:**

I'm unsure of the cause, but somehow a version of node has been cached in the wrong directory. 
The only way to fix this problem is to delete the cache manually if you have access to the runner.

**Action version:**
v3

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [ ] Hosted
- [x] Self-hosted

**Tools version:**
Requesting v16.15.1 gave v14.16.1 (requested cache path was correct, but the version of node inside was wrong)

**Repro steps:**  
I do not know how to reproduce this with normal usage, but to test on a self hosted runner, install a version of node into the wrong directory.

**Expected behavior:**
A version of node requested should be installed

**Actual behavior:**
A corrupted cached version is provided

dmitry-shibanov:
Hello @ben-styling. Thank you for your report. Could you please provide simple steps to reproduce the issue or debug logs ? Which version of runner do you use ?

ben-styling:
Hi @dmitry-shibanov, thank you for your reply!

I'm very sorry, but I don't know how to replicate the issue we're having without manually putting the wrong binary in the wrong cache directory.

There are not reproduction steps, but while working on a PR for this, I ran the action on a self-hosted runner to install v16.16.0
Then in `_work/_tool/node` I copied 16.16.0 to 16.15.1 `cp -Ra ./16.16.0 ./16.15.1`
When running the action to request v16.15.1, the actual version was 16.16.0.

I'm not sure what you mean when asking which version of runner we use. I know that our runners are x64 linux, but I think once the cache is corrupt, the behaviour is the same on all platforms.

I think these are the logs you're asking for, let me know if not!

```
##[debug]Evaluating condition for step: 'Setup node.js'
##[debug]Evaluating: success()
##[debug]Evaluating success:
##[debug]=> true
##[debug]Result: true
##[debug]Starting: Setup node.js
##[debug]Register post job cleanup for action: actions/setup-node@v3
##[debug]Loading inputs
##[debug]Evaluating: github.token
##[debug]Evaluating Index:
##[debug]..Evaluating github:
##[debug]..=> Object
##[debug]..Evaluating String:
##[debug]..=> 'token'
##[debug]=> '***'
##[debug]Result: '***'
##[debug]Loading env
Run actions/setup-node@v3
  with:
    node-version-file: .nvmrc
    check-latest: true
[...]

PR #542, opened 2022-07-12
Fix: validate cached version

ben-styling:
**Description:**

Validates the cached binary version against the versionSpec. If it doesn't match, then try to re-download.

**Related issue:**
[Related issue](https://github.com/actions/setup-node/issues/541)

**Check list:**

### DIFF
--- a/__tests__/installer.test.ts
+++ b/__tests__/installer.test.ts
@@ -109,7 +109,7 @@ describe('setup-node', () => {
 
     // @actions/exec
     getExecOutputSpy = jest.spyOn(exec, 'getExecOutput');
-    getExecOutputSpy.mockImplementation(() => 'v16.15.0');
+    getExecOutputSpy.mockImplementation(() => ({stdout: 'v12.16.1'}));
   });
 
   afterEach(() => {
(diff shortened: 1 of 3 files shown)


vsafonkin:
Hello @ben-styling, tool-cache `find` method already uses a `semver` satisfy check: https://github.com/actions/toolkit/blob/main/packages/tool-cache/src/tool-cache.ts#L727

As I see, your changes just duplicate this check, I'm not sure if this will solve your issue.

ben-styling:
Hi @vsafonkin, thank you for taking a look at this!

Apologies for my naivety when working on this, it's my first time delving into how actions work!

You're right, I was calling `node --version` before adding the toolpath to PATH. I've updated this PR to add the toolpath PATH before calling `node --version`.

As far as I cal tell, toolcache is interring the version of a tool based on the toolpath. 

This check ensures the binary is the correct version, not just the toolpath.

I'm yet to discover how a tool could end up in the wrong directory. I'm currently discussing this with @dmitry-shibanov on [this issue](https://github.com/actions/setup-node/issues/541)

ben-styling:
Thank you for that link, it's a huge help to my understanding!

Based on this, I think you're right, detection of previous installation errors is beyond the scope of `setup-*` actions. 

I'll come back here/on the issue, and post updates if I find out how our runners are experiencing this.

Closing this PR.

PR #546, opened 2022-07-17
Add Option to Enable corepack

### LINKED ISSUE #531: Corepack Support
**Description:**
Support using https://github.com/nodejs/corepack to manage non-NPM package managers.

Ideally in between installing node and bootstrapping the package manager cache, this action:

* enables corepack (could be a no-op depending on the node version)
* caches/restores the corepack root
* runs `corepack prepare --active`

(after some research, that last one may not necessary but instead makes an implicit step explicit)

At least MVP would only allow getting the version to restore from the root's `packageManager` field.

This behavior could either be trigged by the presence of the `packageManager` field in the root `package.json` (might be suprising), `cache: 'auto'` as possibly envisioned in #306, or `corepack: true` 

**Justification:**
Both pnpm and yarn support corepack, and yarn recommendeds corepack for package manager versioning. Currently using corepack with this action will break when using the `cache` key as this action assumes the appropriate version manager has already been installed. However, you don't necessarily want to bootstrap with corepack until the appropriate node version has been installed and configured.

Related: #530 #182 

Yes

BeeeQueue:
**Description:**
Adds the `corepack` option, which if `true` will run `corepack enable` to enable corepack.

You can see a test run of it here: https://github.com/BeeeQueue/arm-server/runs/7380286080?check_suite_focus=true

**Related issue:**
Closes #531 
#480 - @dmitry-shibanov says that this change would not bring enough value to users, but I agree with @shellscape that being able to move most if not all of the node setup into one step is quite nice 🙂 
#482 - That PR adds automatic support for it which could potentially be breaking, this PR adds an option which should not break anything

**Check list:**


BeeeQueue:
I can't figure out why the test is failing based on the other tests... 😕

wuzzeb:
Is it possible to completely remove pnpm/action-setup with this?  What I mean is

```yaml
- uses: actions/node-setup@v3
  with:
    node-version: "18"
    corepack: true
    cache: "pnpm"
```

would be enough with no other actions so no pnpm installed at all before node-setup?  I would think so because node is installed, corepack enable is run, and then after that the caching is setup.  Thus the first time node-setup calls pnpm for the caching, corepack should install pnpm.

If so, I think the documentation in `advanced-usage.md` should be updated, perhaps add a section to the caching showing just this configuration without pnpm/action-setup.

__tests__/installer.test.ts
```diff
@@ -381,6 +381,21 @@ describe('setup-node', () => {
     }
   }, 100000);
 
+  it('enables corepack if specified', async () => {
+    os.platform = 'linux';
+    os.arch = 'x64';
+
+    inputs['node-version'] = '12';
```
styfle, line 388: Corepack doesn’t work with Node.js 12

action.yml
```diff
@@ -13,7 +13,10 @@ inputs:
     description: 'Target architecture for Node to use. Examples: x86, x64. Will use system architecture by default.'
   check-latest:
     description: 'Set this option if you want the action to check for the latest available version that satisfies the version spec.'
-    default: false
+    default: 'false'
+  corepack:
+    description: 'Set this option if you want to install the core-pack.'
```
styfle, line 18:
```suggestion
    description: 'Set this option if you want to install the corepack.'
```

docs/advanced-usage.md
```diff
@@ -249,3 +268,31 @@ steps:
...
+
+`corepack` is a new experimental feature that automatically installs package managers as they are used. [Read more about it here](https://nodejs.org/docs/latest/api/corepack.html)
+
+```yaml
+steps:
+- uses: actions/checkout@v3
+- uses: actions/setup-node@v3
+  with:
+    node-version: '14.x'
+    corepack: enable
```
styfle, line 284: Should this be “true”. Or perhaps it should allow arbitrary args to be passed to corepack CLI?

BeeeQueue: I'm so used to writing `corepack enable` in the CLI that I wrote it here too 😅

yarinsa:
@BeeeQueue Thank you for this contribution! When do you think this will get merged? If so?

yarinsa (approved):
Looks good

BeeeQueue:
Presumably after the test starts passing, and I can't get it to on my Windows machine 😢

jtbandes:
@dsame @brcrista Since #531 was [rejected](https://github.com/actions/setup-node/issues/531#issuecomment-1217533976), but it's clear that many people are interested in this feature, could you please provide some feedback on this PR and/or #482 which are two proposals to add corepack support? Currently it seems the PRs have received no feedback from maintainers.

docs/advanced-usage.md
```diff
@@ -269,3 +288,31 @@ steps:
   run: yarn install --immutable
 ```
 NOTE: As per https://github.com/actions/setup-node/issues/49 you cannot use `secrets.GITHUB_TOKEN` to access private GitHub Packages within the same organisation but in a different repository.
+
+## Automatically enable `corepack`
+
+_Requires Node >=14.x_
```
styfle, line 294:
```suggestion
_Requires Node >=14.19.0
```

src/main.ts
```diff
@@ -58,6 +58,11 @@ export async function run() {
       auth.configAuthentication(registryUrl, alwaysAuth);
     }
 
+    const enableCorepack = core.getBooleanInput('corepack');
+    if (enableCorepack) {
+      await exec.exec('corepack', ['enable']);
```
styfle, line 63: Note this will work with pnpm and yarn, but it wont work with npm which requires `corepack enable npm` explicitly.

styfle: One option is to change the configuration from `corepack: <boolean>` to `corpack: string[]` so that the user can pass an array of args. Then it could look like `corepack: ['enable', 'npm']`

The other solution is to do the same thing that PR #482 is doing and read `packageManager` config to set the correct value.

BeeeQueue: IMO the best solution would be to accept `true | <package manager>` and add the value to the args if it isn't `true`

styfle (approved):
🎉

yarinsa:
@BeeeQueue What's your ETA on merging this?

brcrista:
@jtbandes it's not rejected; I just said it's not a priority yet: https://github.com/actions/setup-node/issues/531#issuecomment-1218035365

It looks like some of the tests were added are failing.  @BeeeQueue I don't know a lot about this domain, but are you sure `corepack` is installed on the hosted runner?

When I have an issue with an Actions workflow I can't reproduce locally, I usually end up having to add some `console.log` calls to find the problem.

BeeeQueue:
None of the test errors seem to be because of the PR, and would probably pass if they run again which I can't do

BeeeQueue:
Superceded by #651

Issue #529, opened 2022-06-27
Set npm/yarn/pnpm version 


stijn-gravity:
**Description:**
Similar to how we can set the node version I'd like to set the npm version. 
```
node-version: 1.16.x
npm-version: 8.5.5
```

There is a similar issue here https://github.com/actions/setup-node/issues/213, but it's for a different reason and this one can be resolved by setting the node version. However this is not enough for the problems I have been facing.

**Justification:**
Currently I am struggling with a bug in npm from versions 8.11 to 8.13. These are automatically installed when you chose node version 1.16. This means I have to set back my npm version so the ci can run, which I can do in an extra ci step.

However to me the issue is that the npm version will change over time, without our input. We'd like control over this to prevent issues like the one above from showing up.

**Are you willing to submit a PR?**
Not currently but I might in the future.

vsafonkin:
Hi @stijn-gravity, thank you for your proposal, we will consider it.

haneenmahd:
I will be opening a PR on this!

jasikpark:
Two options for specifying a solution could be querying the `volta` section of the package.json if you manage versions via that, or via the `engines` portion of the package.json https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines

kidonng:
With corepack support in #482 you can set yarn/pnpm version.

You can also "set" npm/yarn/pnpm version if you manually install them:

```sh
npm install npm@a.b.c
yarn set version a.b.c
npm install pnpm@a.b.c # Or use pnpm/action-setup which has a version option
```

nickmccurdy:
This should be solved by #531

styfle:
Also see #546

christian-heusel:
Additionally to explicitly setting an NPM version it would be good if this action supported `engines.npm` in `packages.json` (https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) so that it installs the right `npm` version from this.

Otherwise users need to do somewhat ugly workarounds like the one here: https://github.com/christian-heusel/notebooks/commit/790e0e14bdb4a7d3639abb02240f33411591f230

This in turn is needed because only newer versions of `npm` support `min-release-age` which is atleast some help with regard to a lot of the currently ongoing supply chain attacks.

risantos:
[...]

PR #549, opened 2022-07-30
Set package manager versions

haneenmahd:
**Description:**
Added support for specifying package manager versions in the workflow file just as specifying node version.
This only added support for the package managers that are supported by this GitHub action and in this case these are `npm`, `pnpm`, `yarn`.

**Related issue:**
#529 

**Check list:**


haneenmahd:
@gordey4doronin, how do we spec the version of the package manager to be consumed?

PR #610, opened 2022-10-26
feat: add node-lts-codename output

panva:
**Description:**
Adds the installed node's [LTS Codename](https://nodejs.org/api/process.html#processrelease) (or lack of) to the action's output.

**Related issue:**
Don't have an open issue but I find myself in need of this output to determine if a given job should continue-on-error or not irrespective of what my node-version input format was.

**Check list:**

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -32,6 +32,8 @@ outputs:
     description: 'A boolean value to indicate if a cache was hit.'
   node-version:
     description: 'The installed node version.'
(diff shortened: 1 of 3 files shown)


panva:
lts-syntax CI jobs fail because actions/node-versions is not updated to include the current latest (Hydrogen) yet. See https://github.com/actions/node-versions/pull/117#issuecomment-1291648895

Issue #627, opened 2022-11-17
Use restore-key to find cache, when exact match is not available


MatthijsBurgh:
**Description:**
Similar to #410 and #441, but these are closed without any result.
Cache is correctly stored in the `Post Setup Node`, but not hit in the next run. 
```
Run actions/setup-node@v3
  with:
    node-version: 16
    cache: yarn
    always-auth: false
    check-latest: false
    token: ***
Found in cache @ /opt/hostedtoolcache/node/16.18.0/x64
Environment details
  node: v16.18.0
  yarn: 1.22.19
  npm: 8.19.2
  
/usr/local/bin/yarn --version
1.22.19
/usr/local/bin/yarn cache dir
/home/runner/.cache/yarn/v6
yarn cache is not found
```

I think this is a combined bug report and feature request. As I checked the code in this action and it just checks cache with one primary key. Maybe additional `restore-keys` can be provided. This is available in the [JS API](https://github.com/actions/toolkit/blob/819157bf872a49cfcc085190da73894e7091c83c/packages/cache/src/cache.ts#L65).

Maybe `node-cache-${platform}-${packageManager}-` is a good restore-key, which would solve the missing hit with an updated lockfile. This would match the example from the [actions/cache docs](https://github.com/actions/cache/blob/main/examples.md#node---npm)

**Action version:**
`v3`

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [x] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
node 14 & 16, yarn

**Repro steps:**  
public repo: https://github.com/nklayman/vue-cli-plugin-electron-builder

**Expected behavior:**
When lockfile is updated, it should hit the most recent cache in that branch scope.

**Actual behavior:**
When lockfile is updated, it doesn't hit any cache. 
Updated lockfile, no cache hit: https://github.com/nklayman/vue-cli-plugin-electron-builder/actions/runs/3481884289
Same lockfile, cache hit: https://github.com/nklayman/vue-cli-plugin-electron-builder/actions/runs/3487432460

dmitry-shibanov:
Hello @MatthijsBurgh. Could you please rename the issue because it is not a bug but a feature request.

MatthijsBurgh:
> Hello @MatthijsBurgh. Could you please rename the issue because it is not a bug but a feature request.

Done

dsame:
[...]

PR #628, opened 2022-11-17
feat(cache) use restoreKeys as backup to hit cache

MatthijsBurgh:
**Description:**
Add restoreKeys to find cache. When lockfile has changed the primary key will not find the cache. Though using the most recent one in the branch scope should still be an improvement.

~As this is a feature in the actions toolkit, I am not sure additional tests should be added. In case you think it should, maybe it is best for the maintainers to do it.~
I do see that the test need to be extended. Maybe it is best for the maintainers to do it.

**Related issue:**
Fixes https://github.com/actions/setup-node/issues/627

**Check list:**


MatthijsBurgh:
@dmitry-shibanov could you approve the workflow runs? And could you review it?

PR #728, opened 2023-03-26
Replace dash to slash for asdf lts syntax

doong-jo:
**Description:**

> asdf support `lts-*`, but setup-node support `lts/*`
스크린샷 2023-03-28 오전 12 30 09

When using the Node.js LTS code names in asdf and setup-node, there is a difference in the format of the code name. For example, asdf requires the code name with dashes included, while setup-node requires the code name with slashes included. This needs to be corrected.

**Related issue:**
https://github.com/actions/setup-node/pull/373

**Check list:**

### DIFF
--- a/.github/workflows/basic-validation.yml
+++ b/.github/workflows/basic-validation.yml
@@ -15,3 +15,5 @@ jobs:
   call-basic-validation:
     name: Basic validation
     uses: actions/reusable-workflows/.github/workflows/basic-validation.yml@main
+    with:
+      node-version: '20.x'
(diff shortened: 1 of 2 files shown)


dsame (approved):

dusan-trickovic (approved):

dmitry-shibanov:
Hello @doong-jo. Thank you for your pull request. Could you please sync with the main branch and add a new syntax to [versions.yml](https://github.com/actions/setup-node/blob/main/.github/workflows/versions.yml#L38) matrix ?

Issue #488, opened 2022-05-10
`cache: yarn` not working for Yarn 3 (Berry) in a subfolder


kachkaev:
**Description:**

When a repo with Yarn Berry is in a subfolder, `cache: yarn` does not get applied.

**Action version:**

`setup-node@3.1.1`

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**

Yarn 3.2

**Repro steps:**  

Configure a pipeline that checks out the Node repo into a subfolder:

```yaml
    - name: Check out tooling
      uses: actions/checkout@v3
      with:
        path: tooling ## 👈 👀
        repository: example-org/example-tooling-repo
```

Setup Node, as suggested in the action docs:

```yaml
    - name: Setup Node
      uses: actions/setup-node@v3
      with:
        node-version: 18
        cache: yarn
        cache-dependency-path: tooling/yarn.lock
```

**Expected behavior:**

`actions/setup-node` correctly launches the right version of Yarn (which is included into the repo). Therefore, the right folder is cached and the follow-up setup is faster.

**Actual behavior:**

Yarn cache dir is resolved to `/home/runner/.cache/yarn/v6`. This is because `yarn --version` is called in the default directory, which is followed by `yarn cache dir` instead of `yarn config get cacheFolder` inside `tooling`. Cache end up empty and so all packages are re-downloaded again each time.

**Potential solution:**

It’d be nice to pass `working-directory` down to `actions/setup-node` (similar to the `run` task).

**Workaround:**

```diff
     - name: Setup Node
       uses: actions/setup-node@v3
       with:
         node-version: 18
-       cache: yarn
-       cache-dependency-path: tooling/yarn.lock
+        
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      run: echo "dir=$(yarn config get cacheFolder)" >> $GITHUB_OUTPUT
+      shell: bash
+      working-directory: tooling
+
+    - name: Restore yarn cache
+      uses: actions/cache@v3
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: yarn-cache-folder-${{ hashFiles('**/yarn.lock', '.yarnrc.yml') }}
+        restore-keys: |
+          yarn-cache-folder-
```

Cache key is inspired by https://github.com/actions/setup-node/issues/325

[...]

PR #735, opened 2023-04-10
Add project-dir

dsame:
**Description:**

Change current working dir when running package manager command.
 - use new `project-dir` input if set
 - use `path.dirname(core.getInput('cache-dependency-path'))` if `cache-dependency-path` input is set
 - don't change if none above

**Related issue:**
[link to original issue ](https://github.com/actions/setup-node/issues/488)
[link to milestone issue ](https://github.com/actions/runner-images-internal/issues/4965)

**Check list:**

### DIFF
--- a/src/cache-utils.ts
+++ b/src/cache-utils.ts
@@ -1,6 +1,7 @@
 import * as core from '@actions/core';
 import * as exec from '@actions/exec';
 import * as cache from '@actions/cache';
+import path from 'path';
 
 type SupportedPackageManagers = {
   [prop: string]: PackageManagerInfo;
@@ -30,11 +31,14 @@ export const supportedPackageManagers: SupportedPackageManagers = {
   }
 };
 
-export const getCommandOutput = async (toolCommand: string) => {
+export const getCommandOutput = async (
+  toolCommand: string,
+  cwd: string | null
+) => {
   let {stdout, stderr, exitCode} = await exec.getExecOutput(
     toolCommand,
     undefined,
-    {ignoreReturnCode: true}
+    {ignoreReturnCode: true, ...(cwd !== null && {cwd})}
   );
 
   if (exitCode) {
@@ -47,11 +51,29 @@ export const getCommandOutput = async (toolCommand: string) => {
   return stdout.trim();
 };
 
+const getPackageManagerWorkingDir = (): string | null => {
+  const projectDir = core.getInput('project-dir');
+  if (projectDir) {
+    return projectDir;
+  }
+
+  const cache = core.getInput('cache');
+  if (cache !== 'yarn') {
+    return null;
+  }
+
+  const cacheDependencyPath = core.getInput('cache-dependency-path');
+  return cacheDependencyPath ? path.dirname(cacheDependencyPath) : null;
+};
+
 const getPackageManagerVersion = async (
   packageManager: string,
   command: string
 ) => {
-  const stdOut = await getCommandOutput(`${packageManager} ${command}`);
+  const stdOut = await getCommandOutput(
+    `${packageManager} ${command}`,
+    getPackageManagerWorkingDir()
+  );
 
   if (!stdOut) {
     throw new Error(`Could not retrieve version of ${packageManager}`);
@@ -85,7 +107,8 @@ export const getCacheDirectoryPath = async (
   packageManager: string
 ) => {
   const stdOut = await getCommandOutput(
-    packageManagerInfo.getCacheFolderCommand
+    packageManagerInfo.getCacheFolderCommand,
+    getPackageManagerWorkingDir()
   );
 
   if (!stdOut) {
--- a/action.yml
+++ b/action.yml
@@ -25,7 +25,9 @@ inputs:
     description: 'Used to specify a package manager for caching in the default directory. Supported values: npm, yarn, pnpm.'
   cache-dependency-path:
     description: 'Used to specify the path to a dependency file: package-lock.json, yarn.lock, etc. Supports wildcards or a list of file names for caching multiple dependencies.'
-# TODO: add input to control forcing to pull from cloud or dist. 
+  project-dir:
+    description: 'Optional path used as a current working directory during executing package manager commands. Important if project directory is not the same as root.'
+# TODO: add input to control forcing to pull from cloud or dist.
 #       escape valve for someone having issues or needing the absolute latest which isn't cached yet
 outputs:
   cache-hit: 
--- a/dist/cache-save/index.js
+++ b/dist/cache-save/index.js
@@ -61136,11 +61136,15 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
         step((generator = generator.apply(thisArg, _arguments || [])).next());
     });
 };
+var __importDefault = (this && this.__importDefault) || function (mod) {
+    return (mod && mod.__esModule) ? mod : { "default": mod };
+};
 Object.defineProperty(exports, "__esModule", ({ value: true }));
 exports.isCacheFeatureAvailable = exports.isGhes = exports.getCacheDirectoryPath = exports.getPackageManagerInfo = exports.getCommandOutput = exports.supportedPackageManagers = void 0;
 const core = __importStar(__nccwpck_require__(2186));
 const exec = __importStar(__nccwpck_require__(1514));
 const cache = __importStar(__nccwpck_require__(7799));
+const path_1 = __importDefault(__nccwpck_require__(1017));
 exports.supportedPackageManagers = {
     npm: {
         lockFilePatterns: ['package-lock.json', 'npm-shrinkwrap.json', 'yarn.lock'],
@@ -61159,8 +61163,8 @@ exports.supportedPackageManagers = {
         getCacheFolderCommand: 'yarn config get cacheFolder'
     }
 };
-const getCommandOutput = (toolCommand) => __awaiter(void 0, void 0, void 0, function* () {
-    let { stdout, stderr, exitCode } = yield exec.getExecOutput(toolCommand, undefined, { ignoreReturnCode: true });
+const getCommandOutput = (toolCommand, cwd) => __awaiter(void 0, void 0, void 0, function* () {
+    let { stdout, stderr, exitCode } = yield exec.getExecOutput(toolCommand, undefined, Object.assign({ ignoreReturnCode: true }, (cwd !== null && { cwd })));
     if (exitCode) {
         stderr = !stderr.trim()
             ? `The '${toolCommand}' command failed with exit code: ${exitCode}`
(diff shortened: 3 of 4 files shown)


kish3499 (commented):
hello

src/cache-utils.ts
```diff
@@ -47,11 +51,24 @@ export const getCommandOutput = async (toolCommand: string) => {
   return stdout.trim();
 };
 
-const getPackageManagerVersion = async (
+export const getPackageManagerWorkingDir = (): string | null => {
+  const cache = core.getInput('cache');
+  if (cache !== 'yarn') {
+    return null;
+  }
```
dmitry-shibanov, line 58: I think we should not do it for yarn specifically. We can remove it.

src/cache-utils.ts
```diff
@@ -47,11 +51,24 @@ export const getCommandOutput = async (toolCommand: string) => {
...
+
+  const cacheDependencyPath = core.getInput('cache-dependency-path');
+  return cacheDependencyPath ? path.dirname(cacheDependencyPath) : null;
```
dmitry-shibanov, line 61: we can change return default return argument to undefined.

src/cache-utils.ts
```diff
@@ -47,11 +51,24 @@ export const getCommandOutput = async (toolCommand: string) => {
...
+};
+
+export const getPackageManagerVersion = async (
   packageManager: string,
   command: string
 ) => {
-  const stdOut = await getCommandOutput(`${packageManager} ${command}`);
+  const stdOut = await getCommandOutput(
+    `${packageManager} ${command}`,
+    getPackageManagerWorkingDir()
```
dmitry-shibanov, line 70: I think we can call `getPackageManagerWorkingDir` directly in getCommandOutput.

src/cache-utils.ts
dsame: It is too dangerous, unpredictable and was not requested. Yarn requires `cd` to the subdir in order to fix the problem with the incorrect version of yarn, i am not sure how the other package managers will be affected by forcing the to run in the other dir than root `package.json`

If we want to let other project managers to be run in not-root directory we should add `project-dir` input. Setting it explicitly the users accept the possible negative consequences.

src/cache-utils.ts
dsame: the function returning undefined suspected to be void, and `undefined` means "unassigned" while i want to assign and return meaningful value "no path"

src/cache-utils.ts
dsame: it would happen a pain some day to use `getCommandOutput` for other purposes. [I wrapped it](https://github.com/actions/setup-node/pull/735/commits/352719599292b2ba43c5e7a43e354ac8c2cdf395#diff-b602b70ff15dc447863bdb26a26be6c2ef6c8aa51314213a26f1e0f162618abfR64`) into `getPackageManagerCommandOutput` instead

rllane17 (approved):

rllane17:
a

src/cache-utils.ts
rllane17: leave it

rllane17 (approved):

src/cache-utils.ts
```diff
@@ -66,37 +107,191 @@ export const getPackageManagerInfo = async (packageManager: string) => {
...
+/**
+ * Expands (converts) the string input `cache-dependency-path` to list of directories that
+ * may be project roots
+ * @param cacheDependencyPath - either a single string or multiline string with possible glob patterns
+ *                              expected to be the result of `core.getInput('cache-dependency-path')`
+ * @return list of directories and possible
+ */
+const cacheDependencyPathToProjectsDirectories = async (
+  cacheDependencyPath: string
+): Promise<string[]> => {
```
IvanZosimov, line 189:
```suggestion
const getProjectDirectories = async (
  cacheDependencyPath: string
): Promise<string[]> => {
```

IvanZosimov: I'd completely remove function `expandCacheDependencyPath()` because I believe it's not needed. Why? Because function `getInput()` always returns a string. And this string will be well formatted to be supplied to the `glob.create()` right away.
**Example:**
```yaml
- uses: actions/setup-node@v3
   with:
      cache-dependency-path: |
         path1
         path2
         path3
```
If we check the `cacheDependencyPath` input data it will look like: "path1\npath2\npath3".

So I'd use `glob` package right away to find all files matched by the `cacheDependecyPath`. Next I'd convert the list of paths to lock files, to list of paths to directories which contain these lock files. Then, to make this array unique, I'd convert it to Set and then return back Array made from Set using Array.from().

IvanZosimov (changes requested):
P.S. I believe we need to decrease number of functions and merge some of them into one, as it's very hard to jump from one to another and don't loose the thread of logic.

src/cache-utils.ts
dsame: The only point is using Set implies memory allocation for binary tree that is too expensive for the small number of elements, the direct search is used instead

src/cache-utils.ts
```diff
@@ -66,35 +103,131 @@ export const getPackageManagerInfo = async (packageManager: string) => {
...
+      .then(globber => globber.glob())) || [''];
+    memoizedCacheDependencies[cacheDependencyPath] = cacheDependenciesPaths;
+  }
+
+  const existingDirectories: string[] = cacheDependenciesPaths
+    .map(cacheDependencyPath => path.dirname(cacheDependencyPath))
+    // uniq in order to do not traverse the same directories during the further processing
+    .filter(
+      (cachePath, i, result) =>
+        cachePath != null && result.indexOf(cachePath) === i
```
nikolai-laevskii, line 145: Scanning through `result` like that on each iteration would have O(n^2) time complexity which is very unnecessary here and uniqueness filter is general-purpose enough to deserve it's own declaration. Consider:
```typescript
// somwhere in utils folder or something
const unique = () => {
  const encountered = new Set();
  return (value) => {
    if (encountered.has(value)) return false;
    encountered.add(value);
    return true;
  }
}

/* ... */

// here
.filter(cachePath => cachePath !== null)
.filter(unique())

/* ... */
```

nikolai-laevskii (changes requested):

src/cache-utils.ts
```diff
@@ -66,35 +103,131 @@ export const getPackageManagerInfo = async (packageManager: string) => {
...
+const getProjectDirectoriesFromCacheDependencyPath = async (
+  cacheDependencyPath: string
+): Promise<string[]> => {
+  let cacheDependenciesPaths: string[];
+
+  // memoize unglobbed paths to avoid traversing FS
+  const memoized = memoizedCacheDependencies[cacheDependencyPath];
+  if (memoized) {
+    cacheDependenciesPaths = memoized;
+  } else {
```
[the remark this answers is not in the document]
dsame: @nikolai-laevskii i love your approach and it was my starting point but during the reviewing it was request to inline the functions in order to have all the code under the fingers. With the small amount of the total codebase it makes some sense to do not add extra functions.

i remove `.then` in the case, yes, it does not worth to be here

src/cache-utils.ts
```diff
@@ -66,35 +103,131 @@ export const getPackageManagerInfo = async (packageManager: string) => {
...
+ * Expands (converts) the string input `cache-dependency-path` to list of directories that
+ * may be project roots
+ * @param cacheDependencyPath - either a single string or multiline string with possible glob patterns
+ *                              expected to be the result of `core.getInput('cache-dependency-path')`
+ * @return list of directories and possible
+ */
+const getProjectDirectoriesFromCacheDependencyPath = async (
+  cacheDependencyPath: string
+): Promise<string[]> => {
+  let cacheDependenciesPaths: string[];
```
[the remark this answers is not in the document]
dsame: plural/singular is distinctive and reflects the purpose of this variable. I see ending `s` as well noticeable naming convention as soon as one get used to.

src/cache-utils.ts
```diff
@@ -66,35 +103,131 @@ export const getPackageManagerInfo = async (packageManager: string) => {
...
+    cacheDependenciesPaths = memoized;
+  } else {
+    cacheDependenciesPaths = (await glob
+      .create(cacheDependencyPath)
+      .then(globber => globber.glob())) || [''];
+    memoizedCacheDependencies[cacheDependencyPath] = cacheDependenciesPaths;
+  }
+
+  const existingDirectories: string[] = cacheDependenciesPaths
+    .map(cacheDependencyPath => path.dirname(cacheDependencyPath))
```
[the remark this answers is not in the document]
dsame: there 3 kind of "paths" through the code: paths of cache dependencies, paths of the projects related to them and paths of cached directories. I intentionally added verbosity on every use of "a path" to do not be lost.

But in this particular case the only paths in the scope are dependency paths and with the simplification `.map(path.dirname)` it worth to get rid off extra qualificatiors, i agree

src/cache-utils.ts
dsame: We know the array to be iterated in this particular case always will be very tiny - 2 or 3, well max 5 items and iteration over them is chipper or equal than allocation/freeing of the not so simple object and building the binary tree for just few values.

src/cache-utils.ts
```diff
@@ -66,35 +103,131 @@ export const getPackageManagerInfo = async (packageManager: string) => {
...
-  packageManager: string
-) => {
-  const stdOut = await getCommandOutput(
-    packageManagerInfo.getCacheFolderCommand
+  cacheDependencyPath: string
+): Promise<string[]> => {
+  const projectDirectories = await getProjectDirectoriesFromCacheDependencyPath(
+    cacheDependencyPath
   );
+  const cacheFoldersPaths = await Promise.all(
```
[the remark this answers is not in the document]
dsame: Yes, i would have done it in this way but once again - "inlining is preferable"

Please keep in mind, this PR likely to be a temporary solution because the inherited approach  "one PM per repo" is not true for "multiple directory repo" and should be re-thought from scratch. This is why i prefer to stay with the accustomed coding style.

nikolai-laevskii (approved):

.github/workflows/e2e-cache.yml
```diff
@@ -134,3 +134,78 @@ jobs:
...
+        env:
+          YARN_ENABLE_IMMUTABLE_INSTALLS: false
+        run: |
+          rm package.json
+          rm package-lock.json
+          echo "create yarn2 project in the sub2"
+          mkdir sub2
+          cd sub2
+          cat <<EOT >package.json
+          {
```
MaksimZhukov, line 158: Can we move the step content into a separate sh script?

__tests__/cache-save.test.ts
```diff
@@ -127,18 +129,28 @@ describe('run', () => {
...
+      expect(getCommandOutputSpy).toHaveBeenCalledTimes(0);
+      expect(debugSpy).toHaveBeenCalledTimes(0);
+      /*
+      expect(debugSpy).toHaveBeenCalledWith(
+        'Consumed yarn version is 2.2.3 (working dir: "")'
+      );
+      expect(debugSpy).toHaveBeenCalledWith(
+        'yarn\'s cache folder "/some/random/path/yarn2" configured for the root directory'
+      );
+       */
```
MaksimZhukov, line 153: Why do we need these comments?

IvanZosimov (approved):

marko-zivic-93 (approved):
LGTM! :)

Issue #325, opened 2021-09-03
Performance opportunity for yarn 3 cache


belgattitude:
Thanks for the new cache feature. Much easier.

After few weeks, I realized that while it supports yarn, there's some improvements that can be made.

Yarn 3 (probably yarn 2+ too) manages downloaded archives pretty well (`.yarn/cache/*.zip`) and invalidating on yarn.lock changes does not take that into account and I saw a lot of cache misses.

As an example I converted back to action-cache to illustrate and test.

The P/R is here https://github.com/belgattitude/nextjs-monorepo-example/pull/369.

I'm wondering if a similar approach could be done with setup-node ?

PS: Key points

<details>
  <summary>Example with action cache</summary>

```
      - name: Get yarn cache directory path
        id: yarn-cache-dir-path
        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
      - name: Restore yarn cache
        uses: actions/cache@v2
        id: yarn-cache 
        with:
          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}       
          key: yarn-cache-folder-os-${{ runner.os }}-node-${{ env.node-version }}-${{ hashFiles('**/yarn.lock', '.yarnrc.yml') }}      
          restore-keys: |
            yarn-cache-folder-os-${{ runner.os }}-node-${{ env.node-version }}-
            yarn-cache-folder-os-${{ runner.os }}-
```
</details>

<details>
  <summary>Result after a new package was added</summary>

```
yarn install --immutable
➤ YN0000: ┌ Fetch step
  ➤ YN0013: │ 1719 packages were already cached, one had to be fetched (superjson@npm:1.7.5)
➤ YN0000: └ Completed in 1s 124ms

Cache Size: ~127 MB (133261279 B)
Cache saved successfully
Cache saved with key: yarn-cache-folder-os-Linux-node--f118ea4bee07eada9df36ad2e83fd6febcbaf06b5b8962689c7650659e872ad3
```
</details>

I

merceyz:
It can be even more efficient by not including the OS and Node version in the cache key as well - https://github.com/actions/setup-node/pull/272#issuecomment-873564091

nickmccurdy:
This only applies to zero install, right? In that case there probably needs to be a way to detect or toggle it.

dmitry-shibanov:
[...]

PR #744, opened 2023-04-24
Do not ivalidate the cache entirely on lock file change

dsame:
**Description:**

Add a function `cache-utils.ts:repoHasYarn3ManagedCache` returns `true` if there's at least one project with yarn3 managed cache

Yarn3 managed cache enables safe use of the previous cache `cache.restoreCache(cachePaths, primaryKey, [keyPrefix])`
because obsolete dependences are removed by Yarn not letting cache to grow endlessly

**Related issue:**
https://github.com/actions/setup-node/issues/325

The PR initially contained Node version in the key but it was removed due to
https://github.com/actions/setup-node/issues/325#issuecomment-1378444559

**Check list:**

### DIFF
--- a/src/cache-restore.ts
+++ b/src/cache-restore.ts
@@ -8,6 +8,7 @@ import {State} from './constants';
 import {
   getCacheDirectories,
   getPackageManagerInfo,
+  repoHasYarn3ManagedCache,
   PackageManagerInfo
 } from './cache-utils';
 
@@ -37,12 +38,16 @@ export const restoreCache = async (
     );
   }
 
-  const primaryKey = `node-cache-${platform}-${packageManager}-${fileHash}`;
+  const keyPrefix = `node-cache-${platform}-${packageManager}`;
+  const primaryKey = `${keyPrefix}-${fileHash}`;
   core.debug(`primary key is ${primaryKey}`);
 
   core.saveState(State.CachePrimaryKey, primaryKey);
 
-  const cacheKey = await cache.restoreCache(cachePaths, primaryKey);
+  const cacheKey = (await repoHasYarn3ManagedCache(packageManagerInfo))
+    ? await cache.restoreCache(cachePaths, primaryKey, [keyPrefix])
+    : await cache.restoreCache(cachePaths, primaryKey);
+
   core.setOutput('cache-hit', Boolean(cacheKey));
 
   if (!cacheKey) {
--- a/src/cache-restore.ts
+++ b/src/cache-restore.ts
@@ -8,7 +8,7 @@ import {State} from './constants';
 import {
   getCacheDirectories,
   getPackageManagerInfo,
-  repoHasYarn3ManagedCache,
+  repoHasYarnBerryManagedDependencies,
   PackageManagerInfo
 } from './cache-utils';
 
@@ -44,12 +44,19 @@ export const restoreCache = async (
 
   core.saveState(State.CachePrimaryKey, primaryKey);
 
-  const cacheKey = (await repoHasYarn3ManagedCache(
+  const isManagedByYarnBerry = await repoHasYarnBerryManagedDependencies(
     packageManagerInfo,
     cacheDependencyPath
-  ))
-    ? await cache.restoreCache(cachePaths, primaryKey, [keyPrefix])
-    : await cache.restoreCache(cachePaths, primaryKey);
+  );
+  let cacheKey: string | undefined;
+  if (isManagedByYarnBerry) {
+    core.info(
+      'All dependencies are managed locally by yarn3, the previous cache can be used'
+    );
+    cacheKey = await cache.restoreCache(cachePaths, primaryKey, [keyPrefix]);
+  } else {
+    cacheKey = await cache.restoreCache(cachePaths, primaryKey);
+  }
 
   core.setOutput('cache-hit', Boolean(cacheKey));
 
--- a/src/cache-restore.ts
+++ b/src/cache-restore.ts
@@ -44,7 +44,10 @@ export const restoreCache = async (
 
   core.saveState(State.CachePrimaryKey, primaryKey);
 
-  const cacheKey = (await repoHasYarn3ManagedCache(packageManagerInfo))
+  const cacheKey = (await repoHasYarn3ManagedCache(
+    packageManagerInfo,
+    cacheDependencyPath
+  ))
     ? await cache.restoreCache(cachePaths, primaryKey, [keyPrefix])
     : await cache.restoreCache(cachePaths, primaryKey);
 
--- a/src/cache-utils.ts
+++ b/src/cache-utils.ts
@@ -111,6 +111,20 @@ export const getPackageManagerInfo = async (packageManager: string) => {
 };
 
 /**
+ * getProjectDirectoriesFromCacheDependencyPath is called twice during `restoreCache`
+ *  - first through `getCacheDirectories`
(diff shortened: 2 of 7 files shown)


marko-zivic-93 (approved):
From the code standpoint it looks ok to me. I am not sure about the semantics :)

src/cache-restore.ts
```diff
@@ -36,12 +39,13 @@ export const restoreCache = async (
     );
   }
 
-  const primaryKey = `node-cache-${platform}-${packageManager}-${fileHash}`;
+  const keyPrefix = `node-cache-${platform}-${packageManager}-${stepId}-${nodeVersion}`;
```
belgattitude, line 42: Curious - Why stepId ?. In my understanding / experience the cache is totally shareable between steps.

nikolai-laevskii (approved):
Looks good to me

src/cache-utils.ts
```diff
@@ -202,6 +221,58 @@ export const getCacheDirectories = async (
...
+): Promise<boolean> => {
+  if (packageManagerInfo.name !== 'yarn') return false;
+
+  const cacheDependencyPath = core.getInput('cache-dependency-path');
+
+  const yarnDirs = cacheDependencyPath
+    ? await getProjectDirectoriesFromCacheDependencyPath(cacheDependencyPath)
+    : [''];
+
+  for (const dir of yarnDirs.length === 0 ? [''] : yarnDirs) {
```
IvanZosimov, line 267:
```suggestion
  for (const dir of !yarnDirs.length ? [''] : yarnDirs) {
```

src/cache-utils.ts
```diff
@@ -110,6 +110,20 @@ export const getPackageManagerInfo = async (packageManager: string) => {
   }
 };
 
+/**
+ * getProjectDirectoriesFromCacheDependencyPath is called twice during `restoreCache`
+ *  - first through `getCacheDirectories`
+ *  - second from `repoHasYarn3ManagedCache`
+ *
+ *  it contains expensive IO operation and thus should be memoized
```
MaksimZhukov, line 118: Is this a typo?
```suggestion
 *  it contains expensive IO operation and thus should be memorised
```

dsame: No, it is by called by intention https://en.wikipedia.org/wiki/Memoization

nikolai-laevskii: @MaksimZhukov nope https://en.wikipedia.org/wiki/Memoization

nikolai-laevskii:
I'll keep my green light. Couldn't find something out of order, still looks good to me. But don't forget to rename branch from `yarn-perfomance` to `yarn-performance` before merging it into actions

marko-zivic-93 (approved):
LGTM!

src/cache-utils.ts
```diff
@@ -202,6 +221,55 @@ export const getCacheDirectories = async (
...
+ *  - if local cache is not explicitly enabled (not yarn3), return false
+ *  - return true otherwise
+ */
+const isCacheManagedByYarn3 = async (directory: string): Promise<boolean> => {
+  const workDir = directory || process.env.GITHUB_WORKSPACE || '.';
+
+  // if .yarn/cache directory exists the cache is managed by version control system
+  const yarnCacheFile = path.join(workDir, '.yarn', 'cache');
+  if (fs.existsSync(yarnCacheFile) && fs.lstatSync(yarnCacheFile).isDirectory())
+    return Promise.resolve(false);
```
IvanZosimov, line 240:
```suggestion
    return false;
```
I'd simplify it as , async function wraps result in Promise by default.

src/cache-utils.ts
```diff
@@ -202,6 +221,55 @@ export const getCacheDirectories = async (
...
+
+  // NOTE: yarn1 returns 'undefined' with rc = 0
```
IvanZosimov, line 242:
```suggestion
  // NOTE: yarn1 returns 'undefined' with return code = 0
```

IvanZosimov: I'd expend `rc` to `return code` to make it more clear.

IvanZosimov (approved):

MaksimZhukov (approved):

src/cache-utils.ts
```diff
@@ -202,6 +222,91 @@ export const getCacheDirectories = async (
...
+    '${enableGlobalCache[12]}
+    '${enableGlobalCache[13]}
+    '${enableGlobalCache[14]}
+    `
+  );
+  // only local cache is not managed by yarn
+  const managed = enableGlobalCache.trim() === 'false';
+  if (managed) {
+    core.debug(`"${workDir}" dependencies are managed by yarn 3 locally`);
+    return Promise.resolve(true);
```
dmitry-shibanov, line 281: we can change `Promise.resolve(true) -> true`

dmitry-shibanov (approved):

belgattitude:
Hey thanks ! a small question

Does it disable the node_modules folder caching when yarn3 zip files are enabled ?

dusan-trickovic (approved):

dsame:
> Does it disable the node_modules folder caching when yarn3 zip files are enabled ?

Hello @belgattitude, the PR does not change the caching strategy in part of node_modules folder, but yarn3 [replaces node_modules with PnP](https://yarnpkg.com/features/pnp) that is supposed to be managed through VCS

arcanis:
Hey - I'm trying to understand this change. Does it mean that the best thing a Yarn project running with this action can do is to NOT use the global cache (`enableGlobalCache: false`) AND to also NOT check-in the `.yarn/cache` folder?

In Yarn v4 we're changing the default to make `enableGlobalCache: true` the default - would it make sense to keep it to `true` when the install is running on a GitHub action?

arcanis:
Ping @dsame ? We'll soon release the 4.0 stable; do we need to automatically set `enableGlobalCache: false` on GH Actions?

belgattitude:
Personally I wouldn't.

The enableGlobalCache new default won't be enough to get the benefits (action cache keys are part of the recipe too). I would let the setup responsability to setup/node or any tool/action/script that help setting up caches.

Applying different defaults depending on env might also create confusion. 

I'd prefer a section in the yarn doc that focuses on a simple recipe for ci and caches (possibly with less emphasis on offline mode or a reference to it as an advanced use case.).  

That would help people implementing cache strategies (azure, vercel... ) to identify quickly what variables are important, what they do and apply a strategy within their "platform" possibilities.

Congrats for Yarn 4.0

MaksimZhukov:
Hello @arcanis!
> I'm trying to understand this change. Does it mean that the best thing a Yarn project running with this action can do is to NOT use the global cache (enableGlobalCache: false) AND to also NOT check-in the .yarn/cache folder?

Yes, if the user uses the local cache and `.yarn/cache` folder is not in the repository, the action will automatically try to restore previous cache even if it is outdated in order to speed up the build.

> In Yarn v4 we're changing the default to make enableGlobalCache: true the default - would it make sense for us to keep it to false when the install is running on a GitHub action, and only there?

Thank you for contacting us and letting us know, we really appreciate it! In our opinion, it is not necessary to leave `enableGlobalCache` set to `false` for GitHub Actions, because this inconsistency can lead to confusion. Users can change this option if they want to use the local cache. We will reflect this behaviour in the action documentation to make it clear.

belgattitude:
@arcanis for context, I realized there was a thread about the new setup node cache.  It starts from there https://github.com/actions/setup-node/issues/325#issuecomment-1535094446

PR #843, opened 2023-08-29
fix: add arch to cached path

pengx17:
**Description:**
The existing node cache key does not take architecture into consideration. However during npm/yarn/etc installing the dependencies may generate different binaries based on the current architecture. Thus if the workflow has matrix that spawn different jobs on different architectures for the same platform may consume the same cache, which might be wrong.

This PR adds the arch to the cached key.

**Related issue:**

**Check list:**


src/cache-restore.ts
```diff
@@ -21,6 +21,7 @@ export const restoreCache = async (
     throw new Error(`Caching for '${packageManager}' is not supported`);
   }
   const platform = process.env.RUNNER_OS;
+  const arch = process.env.RUNNER_ARCH;
```
dmitry-shibanov, line 24: I think we can use `os.arch()` or use architecture from [here](https://github.com/actions/setup-node/blob/2a017f350dbf6c4b6bb4508cc83809719115162e/src/main.ts#L22-L35).

pengx17: How about `RUNNER_OS` in line 23? Maybe we should keep the code consistent

HarithaVattikuti: Hi @pengx17, Could you please change the line to use os.arch() or use the arch input. We have to ensure that code works for all type of runners here.

Also please sync your code to use main branch and rebuild the code to pass all the checks.

pengx17: done

yydslaomaque (approved):

yydslaomaque (approved):

vincentbasp:
Hello,
This pull request fixes the issue described in issue by adding the architecture to the cache key. I tested this change on my repository and it works fine.
Do you know when this pr will be merged ?
Thanks to @pengx17  for his work

gowridurgad (approved):

aparnajyothi-y (approved):

aparnajyothi-y:
Hello @pengx17, Could you please update the line os.platform() with process.env.RUNNER_OS to maintain the existing value of process.env.RUNNER_OS for better consistency and clarity with the runner name.

aparnajyothi-y:
Hello @pengx17, a gentle follow up to update the line os.platform() with process.env.RUNNER_OS to maintain the existing value of process.env.RUNNER_OS for better consistency and clarity with the runner name.

pengx17:
> Hello @pengx17, a gentle follow up to update the line os.platform() with process.env.RUNNER_OS to maintain the existing value of process.env.RUNNER_OS for better consistency and clarity with the runner name.

Sorry, just noticed the message. Done

aparnajyothi-y (approved):

Issue #850, opened 2023-09-06
Node 16 deprecation


kinano:
Node16 will be deprecated on 09/11/2023 ([link](https://nodejs.org/en/blog/announcements/nodejs16-eol)). 

Node 16 is used by the action at the moment according to the source code: https://github.com/actions/setup-node/blob/5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d/action.yml#L35-L36

Is there a plan to release a node version update soon?

dusan-trickovic:
Hello, @kinano! Thank you for creating an issue, we'll look into this :)

MikeMcC399:
@dusan-trickovic 

Based on the announcement [GitHub Actions: Transitioning from Node 16 to Node 20](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) the current version of this JavaScript Action [v3.8.1](https://github.com/actions/setup-node/releases/tag/v3.8.1) will cause `node16` deprecation notices to be generated starting on October 23, 2023.

So apart from the fact the Node.js `16` is already in end-of-life status, it will soon start to visibly affect GitHub Action workflows using [actions/setup-node](https://github.com/actions/setup-node/) if no steps are taken to provide an updated release running on `node20`.

Please release a new version supporting `node20` as requested by @kinano!

dusan-trickovic:
Hello @MikeMcC399! Thank you very much for your input, we are looking into this :)

MikeMcC399:
Hello @dusan-trickovic 

> Thank you very much for your input, we are looking into this :)

Do you have any result from looking at the issue? Have you got plans to implement now?

Deprecation notices are supposed to start being displayed today.

MikeMcC399:
- This issue is resolved through PR #866 and release of [actions/setup-node@v4.0.0](https://github.com/actions/setup-node/releases/tag/v4.0.0)

dmitry-shibanov:
Hello everyone. The new version with support of node20 was released in [v4.0.0](https://github.com/actions/setup-node/releases/tag/v4.0.0). For now I'm going to close the issue.

rwader-swi:
@dusan-trickovic sorry to tag but I updated my workflows like `node-version: 20` or even as `'latest'`, and in the logs it downloads the .tar or gets from cache correctly. But on the workflow run's page I still get 'Node 16 deprecated' flooded warnings. Is there anything I'm missing?

MikeMcC399:
[...]

PR #866, opened 2023-10-09
Change node version for action to node20

dmitry-shibanov:
**Description:**
In scope of this pull request we switch node version of the action to node20.
**Related issue:**
https://github.com/actions/setup-node/issues/850

**Check list:**


marko-zivic-93 (approved):

IvanZosimov (approved):

dusan-trickovic (approved):

nikolai-laevskii (approved):

MaksimZhukov (approved):

src/distributions/official_builds/official_builds.ts
```diff
@@ -100,9 +100,9 @@ export default class OfficialBuilds extends BaseDistribution {
             `Received HTTP status code ${err.httpStatusCode}. This usually indicates the rate limit has been exceeded`
           );
         } else {
-          core.info(err.message);
+          core.info((err as Error).message);
```
dsame, line 103: If the compiler does not detect the 'Error' class, shouldn't we dynamically check it, either with a type guard or as simple as (err.message && message || something)?

dmitry-shibanov: I think it is unnecessary because if it was not an error initially, the code related to `core.info` will fail because it accepts only strings. It is unexpected behaviour that string is thrown instead of Error.

MikeMcC399:
@dmitry-shibanov 

- I don't see any update to the [README](https://github.com/actions/setup-node/blob/main/README.md) to list `actions/setup-node@v4`
- At the moment, there is only a tag [v4.0.0](https://github.com/actions/setup-node/tree/v4.0.0).
There is however not yet a tag [v4](https://github.com/actions/setup-node/tree/v4) (nor a branch `v4`). 

Are these points separately planned?

Issue #878, opened 2023-10-24
Slow post-action after v4


TrygveUrdahl:
**Description:**
The post-action step for this action seems to have become much slower after version 4.0.0 was released.
In the two included images, the only major difference should be the version of `actions/setup-node`.

Before:
![image](https://github.com/actions/setup-node/assets/38653456/bafe9670-a139-4c2d-aa66-d9344fcdfc84)

After:
![image](https://github.com/actions/setup-node/assets/38653456/525afa09-fcd5-47ac-b131-ff6cf1f556bd)

**Action version:**
Using `actions/setup-node@v4` 15 hours after 4.0.0 was released.

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
Node 18.18.2, and `Yarn@4.0.0`.

TrygveUrdahl:
I noticed that my workflow files didn't specify what node version to use for this workflow through any `node-version` or `node-version-file`. By adding that, and upgrading my project to use node 20, this issue seems to go away.

nikolai-laevskii:
Thank you for detailed description of the issue! We'll investigate it and come back with details.

panva:
> I noticed that my workflow files didn't specify what node version to use for this workflow through any `node-version` or `node-version-file`. By adding that, and upgrading my project to use node 20, this issue seems to go away.

My workflows do always specify `node-version` but the post-action is slow as well.

I also have a flow log [here](https://github.com/panva/jose/actions/runs/6643396154/job/18050359751#step:3:10) which despite requesting lts/iron resolved `20.8.1` which isn't lts/iron. This particular setup-node use is from a reused worflow, when setup-node is used outside of a reused workflow the correct `20.9.0` is resolved.

bodinsamuel:
Just want to report the same thing, from 5s -> 2m15s (Node 18)
Previous: https://github.com/specfy/stack-analyser/actions/runs/6664890448/job/18113374316
With the upgrade: https://github.com/specfy/stack-analyser/actions/runs/6664871105/job/18113316689?pr=84

ledermann:
Same here in multiple projects using JS with Yarn. The `Post Setup Node.js` job slows down from a few seconds to about 2:20 minutes after upgrading to v4. 

[...]

PR #907, opened 2023-11-27
Explicitly exit the process to not wait for hanging promises

xiniria:
**Description:**
Since Node.js 19, the default for `keepAlive` was changed from `false` to `true`. As explained in [this similar issue for the Ruby setup action](https://github.com/ruby/setup-ruby/issues/543#issuecomment-1793608370), this causes the Post setup Node action to hang for somewhere between 1 and 5 minutes after running. This PR explicitly exits the process once the `run` function is completed, to not wait for hanging promises.

**Related issue:**
See https://github.com/actions/setup-node/issues/878

**Check list:**

### DIFF
--- a/__tests__/canary-installer.test.ts
+++ b/__tests__/canary-installer.test.ts
@@ -46,6 +46,7 @@ describe('setup-node', () => {
   let isCacheActionAvailable: jest.SpyInstance;
   let getExecOutputSpy: jest.SpyInstance;
   let getJsonSpy: jest.SpyInstance;
+  let processExitSpy: jest.SpyInstance;
 
   beforeEach(() => {
     // @actions/core
@@ -63,6 +64,9 @@ describe('setup-node', () => {
     archSpy = jest.spyOn(osm, 'arch');
     archSpy.mockImplementation(() => os['arch']);
     execSpy = jest.spyOn(cp, 'execSync');
+    processExitSpy = jest
+      .spyOn(process, 'exit')
+      .mockImplementation((() => {}) as () => never);
 
     // @actions/tool-cache
     findSpy = jest.spyOn(tc, 'find');
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -35,6 +35,8 @@ describe('main tests', () => {
 
   let setupNodeJsSpy: jest.SpyInstance;
 
+  let processExitSpy: jest.SpyInstance;
+
   beforeEach(() => {
     inputs = {};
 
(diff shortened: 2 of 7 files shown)


xiniria:
> Change works as a quick fix, but it's worth noting the proper fix would be to merge [actions/toolkit#1572](https://github.com/actions/toolkit/pull/1572) (or a similar fix on actions/toolkit level) otherwise we'd be copy pasting this exit process hack to tons of other `actions/*` that cache internally as they update to Node 20.

I totally agree, but since we have no leverage over when (or even if) this other PR is merged, this should improve things for a lot of users out there in the meantine, don't you think?

Bo98:
Yeah for sure. Though given GitHub control both repositories, there's a chance the right person will see it here.

External contributions to actions/toolkit are not currently being reviewed (and haven't for a year) so it felt it was worth mentioning in case they want to make a similar fix themselves. If the toolkit and setup-node teams are totally separate then yeah I imagine merging your workaround makes sense.

xiniria:
@nikolai-laevskii could you have a look at the PR?

xaviermarchal:
@nikolai-laevskii any news on this PR that could help us a lot on our delivery?

ehasnain:
It would great to have this simple fix merged. It's not blocking, but somehow nice to have as these ~5 Minutes do some up to a huge number of minutes for teams with large number of developers and multiple builds with every commit.
Thank you @xiniria the PR.

Bo98:
It's already fixed: https://github.com/actions/setup-node/commit/ec97f37504b0cca1fbc763cc0575585d10020c22

And now double-fixed by the `@actions/http-client` update: https://github.com/actions/setup-node/commit/c2ac33f2c62f978d6c944d9648125a294e56dc0b

ehasnain:
Hi @Bo98 for the quick update. Not sure why but we are still experiencing 4-5 Minutes of duration in the post step.

image
image
image

We are using actions/setup-node@v4 with node-version: 20.11 on multiple pipelines.
Maybe here is not the right spot for the bug/issue reporting, but I landed here on going through the past issues and PRs related to the same.

Bo98:
1GB cache is fairly large. Can you confirm the with the "Show timestamps" option that it's actually hanging _after_ the `Cached saved with key` line?

ehasnain:
Hi @Bo98, your guess is right. The size of 1 GB is actually taking quite long. The step ends right after `Cache saved with key: ...` is logged.
image

Although a compression of similar data (size and number of files), run manually on the same server and disk takes just seconds. I'll try to dig in further. Thank you already for looking into this.

lalkalol1907 (approved):

blackakula:
Hi @ehasnain,
Did you solve the problem?

Do you use a self-hosted runner? The cache by default is saved in the runner instance filesystem. The runner instance (depending on implementation) could be replaced/launched on demand. For example, launching EC2 instance from a template by the load balancer. And all previously generated cache does not exist in a new fresh runner instance.
This problem is not related to either `setup-node` or any other github action.

Here is the discussion with possible solutions for this problem: https://github.com/orgs/community/discussions/18549#discussioncomment-8431516

v-gowridurgad:
Closing this PR as the change has already been implemented in https://github.com/actions/setup-node/pull/917 and the related issue has also been closed.

Issue #1039, opened 2024-05-09
Support env setting of AGENT_TOOLSDIRECTORY as in the setup-python action 


catscanner:
**Description:**
as in [setup python](https://github.com/actions/setup-python/blob/82c7e631bb3cdc910f68e0081d67478d79c6982d/src/setup-python.ts#L79) 
```
  if (IS_MAC) {
    process.env['AGENT_TOOLSDIRECTORY'] = '/Users/runner/hostedtoolcache';
  }

  if (process.env.AGENT_TOOLSDIRECTORY?.trim()) {
    process.env['RUNNER_TOOL_CACHE'] = process.env['AGENT_TOOLSDIRECTORY'];
  }

  core.debug(
    `Python is expected to be installed into ${process.env['RUNNER_TOOL_CACHE']}`
  );
```
setup-node should read the AGENT_TOOLSDIRECTORY environment variable and if it exists, sets the RUNNER_TOOL_CACHE variable equal to it

A GitHub support agent said:
> I was able to get this to work by forking actions/setup-node and copying the same implementation on actions/setup-python to allow support for AGENT_TOOLSDIRECTORY on my fork - essentially just throwing in those few lines where the run function is defined [here](https://github.com/actions/setup-node/blob/c2ac33f2c62f978d6c944d9648125a294e56dc0b/src/main.ts#L13).

**Justification:**
Other actions already support this declaration and we need to use it to populate a caching system we are building to avoid hitting GitHub.com so often to install programming languages.

**Are you willing to submit a PR?**
possibly, I will need to check schedule with our engineering team.

aparnajyothi-y:
Hello @catscanner, Thank you for creating this feature request and we will get back to you once we have some feedback on this :)

th-le:
Hi, I opened a PR for this, really appreciate if you could have a look, @aparnajyothi-y

PR #1040, opened 2024-05-10
Support AGENT_TOOLSDIRECTORY as in setup-python action

th-le:
**Description:**

We needs to run a custom workflow to populate tools cache data for self-hosted runners. However it could not override the default downloading location on setup-node action. 

Setup-python already has support for this use case. We'd like to have the same logic applied on this action.

**Related issue:**
[Add link to the related issue.](https://github.com/actions/setup-node/issues/1039)

**Check list:**


snwbr (approved):

snwbr:
it would be good to review this one as it's a small yet an important one

CosmicJesterX (approved):

CosmicJesterX (approved):

PR #1172, opened 2024-12-14
Added Node version to cache key

### LINKED ISSUE #1171: Include Node.js version in cache key
**Description:**
The Node.js version should be included in the cache key. The major version number, (e.g., 22 in 22.12) should be used to avoid unnecessary invalidation when upgrading to patch or minor versions.

This is similar to https://github.com/actions/setup-node/issues/1008.

**Justification:**
Some dependencies need to be rebuilt for specific Node versions. Reusing the cache for an older version results in dependency installation failures.

Yes.

clintonb:
**Description:**
The Node.js version is now included in the cache key to ensure caches are invalidated, and dependencies reinstalled, when changing Node.js versions.

**Related issue:**
Fixes #1171 

**Check list:**

### DIFF
--- a/__tests__/cache-restore.test.ts
+++ b/__tests__/cache-restore.test.ts
@@ -115,7 +115,7 @@ describe('cache-restore', () => {
     it.each([['npm7'], ['npm6'], ['pnpm6'], ['yarn1'], ['yarn2'], ['random']])(
       'Throw an error because %s is not supported',
       async packageManager => {
-        await expect(restoreCache(packageManager, '')).rejects.toThrow(
+        await expect(restoreCache(packageManager, '', '')).rejects.toThrow(
           `Caching for '${packageManager}' is not supported`
         );
       }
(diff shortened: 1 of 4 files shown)


Furka-090 (approved):

Lmillan123 (approved):

textbook:
Shouldn't this use the actual major Node version in the cache key? The linked issue says:

> The major version number, (e.g., 22 in 22.12) should be used to avoid unnecessary invalidation when upgrading to patch or minor versions.

That makes sense to me. But the implementation here uses `version`, which is either `core.getInput("node-version")` (any [supported version](https://github.com/actions/setup-node?tab=readme-ov-file#supported-version-syntax)) or whatever was pulled from the `node-version-file`. This _could_ be a major version, but equally could be a minor or patch version, a range, or something like `latest` or `lts/*`. For example any of the following could come from the file:

https://github.com/actions/setup-node/blob/48b90677b6048efbc723b11a94acb950d3f1ac36/__tests__/main.test.ts#L92-L110

I think this should be using the _resolved_ Node version, i.e. what's used as `outputs.node-version`, and taking the major version from it.

PR #1284, opened 2025-04-23
[docs] Add link to the engines.node field definition.

### LINKED ISSUE #1081: docs: Node version file: Clarify effect of version range in engines.node
**Description:**
In `docs/advanced-usage.md#node-version-file`, explain what effect `"engines": { "node": ">=16.0.0" }` would have had if it was not overridden by the volta option.

**Justification:**
I cannot tell from reading whether the example would install node 16, 20, 21 or 22.

Yes.

mk-pmb:
**Description:**
Added a link to the definition of the `engines.node` field in `package.json`.

**Related issue:**
Closes #1081.

**Check list:**

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -72,6 +72,7 @@ steps:
 ```
 
 When using the `package.json` input, the action will look for `volta.node` first. If `volta.node` isn't defined, then it will look for `engines.node`.
+(The latter is defined [here in the npm docs](https://docs.npmjs.com/cli/v10/configuring-npm/package-json/#engines).)
 
 ```json
 {


priyagupta108:
Hi @mk-pmb 👋,
Thanks for the PR! We noticed the reference to the npm documentation regarding the order of checking `volta.node` and `engines.node`, and it seems a bit misleading.

According to the [npm documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-json/#engines), npm checks the `engines.node` field in `package.json` to determine the required Node.js version. On the other hand, `volta.node` is specific to [Volta](https://docs.volta.sh/advanced/workspaces), a separate toolchain manager, and is not part of npm's configuration.

If action prioritizes `volta.node` over `engines.node`, that behavior is specific to action and is not part of standard npm behavior. The current statement could be misinterpreted as describing npm's default behavior, which isn’t accurate.

Looking forward to your thoughts!

mk-pmb:
Good catch! That could indeed be confusing. I updated the branch to insert a new line below the original,

> (The latter is defined [here in the npm docs](https://docs.npmjs.com/cli/v10/configuring-npm/package-json/#engines).)

priyagupta108:
@mk-pmb, Thanks for your contribution! After reviewing the changes and the npm documentation, we feel the npm docs link isn’t necessary in this case.

When using the `package.json` input, the action first checks for `volta.node` and only falls back to `engines.node` if `volta.node` is not defined. This behavior is specific to the action and differs from how `engines.node` is used in npm. Since this behavior is already clearly explained in the `setup-node` documentation, we don’t believe the npm docs link adds significant value.

To keep the documentation focused on the action’s specific purpose, we’ve decided not to include the npm docs link. We’ll be closing the PR now, but please feel free to reach out if you have any concerns or further suggestions.

Issue #1363, opened 2025-09-08
v5 fails with Path Validation Error in Post Run steps


deviantintegral:
**Description:**

Upgrading to v5 causes a a workflow to fail with the following error:

>  Error: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

**Action version:**

5.0.0

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**

Node 22.19.0 / npx.

**Repro steps:**  

- Failing PR: https://github.com/Lullabot/drainpipe/pull/935
- Failing job: https://github.com/Lullabot/drainpipe/actions/runs/17513511413/job/49761979980?pr=935

**Expected behavior:**

No errors!

**Actual behavior:**

The error marks the job as failing.

silas:
Looks like this happens if `(npm|pnpm|yarn|...) install` isn't run and `packageManager` is set.

The work around seems to be to disable [caching](https://github.com/actions/setup-node/blob/a0853c24544627f65ddf259abe73b1d18a591444/README.md#caching-global-packages-data) for jobs that don't run install.

v-priyagupta108:
Hello @deviantintegral 👋,
Thank you for your report. We'll investigate the issue and get back to you with the details!

malice00:
I am having the same issue after the update to v5, however it happens **ONLY** on the windows arm64 branch of my matrix!?

<img width="576" height="978" alt="Image" src="https://github.com/user-attachments/assets/d935c3f5-33e2-4344-a1c1-0be303dd7b9c" />

<img width="2798" height="264" alt="Image" src="https://github.com/user-attachments/assets/a817334d-c628-4869-8c46-3afb6fa4dc32" />

v-gowridurgad:
Hi @deviantintegral, Thank you for reporting this issue and providing the details.
Starting with actions/setup-node@v5, caching is enabled by default when a packageManager field is present in the package.json file, even if no cache inputs have been explicitly set in the workflow. For more information, please refer to this setup-node [documentation](https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data).

In this scenario, caching is automatically enabled. The action will try to cache dependencies, but if the required paths are missing because the dependencies haven’t been installed yet, you may see the message "no cache is being saved." This is expected behaviour when there is nothing available to cache. 

[...]

Issue #1358, opened 2025-09-05
Don't cache automatically in privileged workflows


caugner:
**Description:**
In https://github.com/actions/setup-node/pull/1348, caching was enabled by default if the `package.json` includes the `packageManager` field.

This side-effect is dangerous in privileged workflows with access to secrets and credentials, because it makes them vulnerable to cache poisoning. Code injection in one privileged workflow can be exploited to steal higher value secrets, and credentials in another privileged workflow.

**Action version:**
v5.0.0

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
npm (but issue is not specific)

**Repro steps:**  
1. Create a repo with `package.json` incl. `packageManager` field.
2. Create a `pull_request_target` workflow incl. `actions/setup-node` usage.
 
**Expected behavior:**
Caching should not be enabled by default in privileged workflows.

**Actual behavior:**
Caching is enabled by default.

v-aparnajyothi-y:
Hello @caugner, Thank you for creating this issue. This behavior is expected with [actions/setup-node v5](https://github.com/actions/setup-node/releases/tag/v5.0.0). In this release, caching is automatically enabled if your package.json includes a valid packageManager field (such as pnpm, yarn, or npm). This update is aimed at simplifying dependency caching by default.

To turn off automatic caching and restore the previous behavior, set `package-manager-cache: false` in your `setup-node` step:

```yaml
steps:
- uses: actions/checkout@v5
- uses: actions/setup-node@v5
  with:
    package-manager-cache: false
```
For more information on this update and recommended configuration, please refer to:
[v5.0.0 Release Notes ](https://github.com/actions/setup-node/releases/tag/v5.0.0)
[README – Caching Packages](https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data)

Please feel free to reach us in case of any further calrifications/concerns.

caugner:
Thanks for the quick response, and for confirming the  intended behavior.

I would like to clarify that I'm not asking to restore the old behavior. I'm concerned about the security implications of enabling caching by default **in privileged workflows** (e.g. `pull_request_target` with access to secrets).

As mentioned above, automatic caching creates a vector for **cache poisoning**, where malicious code in one workflow can inject content into the cache, and later exfiltrate secrets or credentials from another privileged job.

[...]

Issue #1357, opened 2025-09-04
v5 fails immediately when using `pnpm`


peterhirn:
**Description:**
I'm using `corepack` to install `pnpm`. Using v5 the action fails immediately, afaict doesn't even make it to the corepack step.

**Action version:**
v5

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
```json
  "engines": {
    "node": "24.7.0"
  },
  "packageManager": "pnpm@10.15.1+sha512.34e538c329b5553014ca8e8f4535997f96180a1d0f614339357449935350d924e22f8614682191264ec33d1462ac21561aff97f6bb18065351c162c7e8f6de67"
```

**Repro steps:**  

```yaml
name: restore

runs:
  using: composite
  steps:
    - name: Install Node.js
      #uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
      uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5
      with:
        node-version-file: package.json

    - name: Install pnpm
      shell: bash
      run: |
        corepack enable
        corepack prepare --activate

    - name: Export pnpm store path
      shell: bash
      run: |
        echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV

    - name: Setup pnpm cache
      uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4
      with:
        path: ${{ env.STORE_PATH }}
        key: ${{ runner.os }}-pnpm-store-${{ hashFiles('pnpm-lock.yaml') }}
        restore-keys: |
          ${{ runner.os }}-pnpm-store-

    - name: Install dependencies
      shell: bash
      run: pnpm install --frozen-lockfile

    - name: Store prune
      shell: bash
      run: pnpm store prune
```

```
Prepare all required actions
Getting action download info
Download action repository 'actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444' (SHA:a0853c24544627f65ddf259abe73b1d18a591444)
Download action repository 'actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809' (SHA:0400d5f644dc74513175e3cd8d07132dd4860809)
Run ./.github/actions/restore
Run actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444
  with:
[...]

Issue #1351, opened 2025-08-26
Action failing since `package-manager-cache` added


glassesUSA:
**Description:**
The addition of `package-manager-cache` broke my workflows with installing PNPM.

**Action version:**
`master`

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [ ] Hosted
- [x] Self-hosted

**Tools version:**
pnpm@10.9.0

**Repro steps:**  
The action fails when trying to install dependencies using `pnpm` with the error `Error: Unable to locate executable file: pnpm. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.`

```
name: Setup workflow
description: |
  This action sets up the environment for the project by installing dependencies and configuring Node.js.

inputs:
  GH_TOKEN:
    description: "GitHub token"
    required: true
  NODE_VERSION:
    description: "Node.js version"
    required: false
    default: "20"

runs:
  using: "composite"
  steps:
    - uses: rtCamp/action-cleanup@master
    - uses: actions/checkout@master
    - uses: actions/setup-node@master
      with:
        node-version: ${{ inputs.NODE_VERSION }}
        registry-url: https://npm.pkg.github.com/
        scope: '@optimaxdev'
    - name: Install pnpm
      uses: pnpm/action-setup@v4

    - name: Use Node.js ${{ inputs.NODE_VERSION }}
      uses: actions/setup-node@v4
      with:
        node-version: ${{ inputs.NODE_VERSION }}
        cache: 'pnpm'

    - name: Install dependencies
      run: pnpm i
      shell: bash
      
```

v-aparnajyothi-y:
Hello @glassesUSA, thank you for creating this issue. I was able to reproduce the problem with actions/setup-node@v4 (without the new package-manager-cache feature) when pnpm was missing in the workflow but cache: pnpm was enabled. Adding pnpm/action-setup@v4 with a specific version resolved it.

Could you please share a public repository where your failure occurs so we can investigate further and confirm. I’ve attached screenshots for reference.

<img width="1414" height="746" alt="Image" src="https://github.com/user-attachments/assets/93a2c7e4-bd8d-4ce5-957f-561a619cedb7" />
<img width="1414" height="746" alt="Image" src="https://github.com/user-attachments/assets/c279d47e-6ae7-40f0-b459-2a6207053ce0" />

[...]

PR #1374, opened 2025-09-22
Update caching defaults for npm and documentation update

priyagupta108:
**Description:**
This PR updates the caching logic to refine the automatic caching feature added in response to Feature Request #686 and implemented in #1348.

- **npm:** Caching is now automatically enabled by default if `package.json` contains a `packageManager` field set to `npm` and no explicit `cache` input is provided.
- **pnpm & yarn:** Caching is disabled by default to avoid compatibility issues. Users who wish to enable caching for `pnpm` or `yarn` can do so manually via the `cache` input.
- **Documentation:** Updated to reflect the new default caching behavior for npm and updated example workflows to use the latest versions.

**Related issues:**

-  #1351 
-  #1357
-  #1358 
-  #1363 

**Check list:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ This action provides the following functionality for GitHub Actions users:
 
 ## Breaking changes in V5 
 
-- Enabled caching by default with package manager detection if no cache input is provided.
-  > For workflows with elevated privileges or access to sensitive information, we recommend disabling automatic caching by setting `package-manager-cache: false` when caching is not needed for secure operation.
+- Enabled caching for npm dependencies by default when the package manager is detected and no cache input is provided.
+  > For workflows with elevated privileges or access to sensitive information, we recommend disabling automatic caching for npm by setting `package-manager-cache: false` when caching is not needed for secure operation.
 
 - Upgraded action from node20 to node24.
   > Make sure your runner is on version v2.327.1 or later to ensure compatibility with this release. [See Release Notes](https://github.com/actions/runner/releases/tag/v2.327.1)
@@ -67,7 +67,8 @@ See [action.yml](action.yml)
     # Default: ''
     cache: ''
 
-    # Used to disable automatic caching based on the package manager field in package.json. By default, caching is enabled if the package manager field is present and no cache input is provided'
+    # Controls automatic caching for npm. By default, caching for npm is enabled if the packageManager field in package.json specifies npm and no explicit cache input is provided.
+    # To disable automatic caching for npm, set package-manager-cache to false.
     # default: true
     package-manager-cache: true
 
@@ -115,7 +116,7 @@ steps:
 - uses: actions/checkout@v5
 - uses: actions/setup-node@v5
   with:
-    node-version: 18
+    node-version: 24
 - run: npm ci
 - run: npm test
 ```
@@ -132,8 +133,8 @@ The `node-version` input supports the Semantic Versioning Specification, for mor
 
 Examples:
 
- - Major versions: `18`, `20`
- - More specific versions: `10.15`, `16.15.1` , `18.4.0`
+ - Major versions: `22`, `24`
+ - More specific versions: `20.19`, `22.17.1` , `24.8.0`
  - NVM LTS syntax: `lts/erbium`, `lts/fermium`, `lts/*`, `lts/-n`
  - Latest release: `*` or `latest`/`current`/`node`
 
@@ -151,18 +152,6 @@ It's **always** recommended to commit the lockfile of your package manager for s
 
 The action has a built-in functionality for caching and restoring dependencies. It uses [actions/cache](https://github.com/actions/cache) under the hood for caching global packages data but requires less configuration settings. Supported package managers are `npm`, `yarn`, `pnpm` (v6.10+). The `cache` input is optional.
 
-Caching is turned on by default when a `packageManager` field is detected in the `package.json` file and no `cache` input is provided. The `package-manager-cache` input provides control over this automatic caching behavior. By default, `package-manager-cache` is set to `true`, which enables caching when a valid package manager field is detected in the `package.json` file. To disable this automatic caching, set the `package-manager-cache` input to `false`.
-
-```yaml
-steps:
-- uses: actions/checkout@v5
-- uses: actions/setup-node@v5
-  with:
-    package-manager-cache: false
-- run: npm ci
-```
-> If no valid `packageManager` field is detected in the `package.json` file, caching will remain disabled unless explicitly configured. For workflows with elevated privileges or access to sensitive information, we recommend disabling automatic caching by setting `package-manager-cache: false` when caching is not needed for secure operation.
-
 The action defaults to search for the dependency file (`package-lock.json`, `npm-shrinkwrap.json` or `yarn.lock`) in the repository root, and uses its hash as a part of the cache key. Use `cache-dependency-path` for cases when multiple dependency files are used, or they are located in different subdirectories.
 
 **Note:** The action does not cache `node_modules`
@@ -176,7 +165,7 @@ steps:
 - uses: actions/checkout@v5
 - uses: actions/setup-node@v5
   with:
-    node-version: 20
+    node-version: 24
     cache: 'npm'
 - run: npm ci
 - run: npm test
@@ -189,12 +178,26 @@ steps:
 - uses: actions/checkout@v5
 - uses: actions/setup-node@v5
   with:
-    node-version: 20
+    node-version: 24
     cache: 'npm'
     cache-dependency-path: subdir/package-lock.json
 - run: npm ci
 - run: npm test
 ```
+
+Caching for npm dependencies is automatically enabled when your `package.json` contains a `packageManager` field set to `npm` and no explicit cache input is provided.
+
+This behavior is controlled by the `package-manager-cache` input, which defaults to `true`. To turn off automatic caching, set `package-manager-cache` to `false`.
+
+```yaml
+steps:
+- uses: actions/checkout@v5
+- uses: actions/setup-node@v5
+  with:
+    package-manager-cache: false
+- run: npm ci
+```
+> If your `package.json` file does not include a `packageManager` field set to `npm`, caching will be disabled unless you explicitly enable it. For workflows with elevated privileges or access to sensitive information, we recommend disabling automatic caching for npm by setting `package-manager-cache: false` when caching is not required for secure operation.
 
 ## Matrix Testing
 
@@ -204,7 +207,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [ 14, 16, 18 ]
+        node: [ 20, 22, 24 ]
     name: Node ${{ matrix.node }} sample
     steps:
       - uses: actions/checkout@v5
@@ -226,7 +229,7 @@ To get a higher rate limit, you can [generate a personal access token on github.
 uses: actions/setup-node@v5
 with:
   token: ${{ secrets.GH_DOTCOM_TOKEN }}
-  node-version: 20
+  node-version: 24
 ```
 
 If the runner is not able to access github.com, any Nodejs versions requested during a workflow run must come from the runner's tool cache. See "[Setting up the tool cache on self-hosted runners without internet access](https://docs.github.com/en/enterprise-server@3.2/admin/github-actions/managing-access-to-actions-from-githubcom/setting-up-the-tool-cache-on-self-hosted-runners-without-internet-access)" for more information.
--- a/__tests__/main.test.ts
+++ b/__tests__/main.test.ts
@@ -288,19 +288,20 @@ describe('main tests', () => {
     it('Should enable caching with the resolved package manager from packageManager field in package.json when the cache input is not provided', async () => {
       inputs['package-manager-cache'] = 'true';
       inputs['cache'] = ''; // No cache input is provided
+      isCacheActionAvailable.mockImplementation(() => true);
 
       inSpy.mockImplementation(name => inputs[name]);
 
       const readFileSpy = jest.spyOn(fs, 'readFileSync');
       readFileSpy.mockImplementation(() =>
         JSON.stringify({
-          packageManager: 'yarn@3.2.0'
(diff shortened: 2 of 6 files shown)


kachkaev:
I believe this change needs to be released as v6. I have already removed explicit caching option for a few projects and they work well because `actions/setup-node@v5` follows `pnpm/action-setup@v4`. If `cache: pnpm` becomes a requirement again in v5, existing workflows will end up burning more CI minutes.

voxpelli:
I would really love if something like my PR can get in as well, been sitting there for 3 years with very little attention: https://github.com/actions/setup-node/pull/702

scottohara:
In the context of `npm` only, I'm interested in whether `packageManager` is the right field to use for this.

My concern stems from the fact that there are multiple ways to specify a package manager and version, and they all slightly differ in their intended purpose:

### `engines.<npm|pnpm|yarn>`
According to [npm's documentation](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines), the `engines` field in `package.json` is described as follows:
> You can specify the version of node that your stuff works on.
> You can also use the "engines" field to specify which versions of npm are capable of properly installing your program.

The key phrasing here seems to be _"....capable of properly installing your program"_.

### `devEngines.packageManager`
According to [npm's documentation](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines), the `devEngines` field in `package.json` is described as follows:
> The devEngines field aids engineers working on a codebase to all be using the same tooling.
> Note: engines and devEngines differ in object shape. They also function very differently. engines is designed to alert the user when a dependency uses a differening npm or node version that the project it's being used in, whereas devEngines is used to alert people interacting with the source code of a project.

The key phrasing here seems to be _"...people interacting with the source code of a project"_.

### packageManager
This field is not defined or documented by `npm`, but is rather [a nodejs concept introduced in v16.16.0](https://nodejs.org/download/release/v16.16.0/docs/api/packages.html#packagemanager) alongside Corepack.

As with Corepack, it is still marked as experimental, and from Node v25, Corepack is expected to stop being distributed by Node.

### Summary
Given all of the above, my take on this is:

❌  `devEngines.packageManager` would be inappropriate to use in this context, as `setup-node` does not exactly fit the definition of _"...people interacting with the source code of a project"_

❌  `packageManager` is an _**experimental**_ field used to support Corepack, and indeed [the docs specifically state that](https://nodejs.org/download/release/v16.16.0/docs/api/corepack.html#supported-package-managers):
>  While npm is a valid option in the `packageManager` property, the lack of shim will cause the global npm to be used.

(so it could be argued that `packageManager: npm` is not really intended for use? It is certainly not listed in [supported package managers](https://nodejs.org/download/release/v16.16.0/docs/api/corepack.html#supported-package-managers).)

✅  `engines.npm` is npm's official way of specifying the version of npm supported, and `setup-node` does seem to fit more with the definition of _"...capable of properly installing your program"_.

Anecdotally, [Heroku notes in their documentation](https://devcenter.heroku.com/articles/nodejs-support#using-npm):
* For specifying npm, use `engines.npm`  (they don't support `packageManager: npm@<version>`)
* For specifying pnpm, use either `packageManager: pnpm@<version>` or `engines.pnpm`
* For specifying yarn, use either `packageManager: yarn@<version>` or `engines.yarn`

My view is that `setup-node` should look in the `engines` field first to determine which package manager is used (based on the presence of `engines.<npm|pnpm|yarn>`), and then fallback to `packageManager`.

I don't think `packageManager: npm` is the right long term choice here, given the fact that Corepack (and by extension, the `packageManager` field itself) is likely to be split out from Node in the future.

That is my 2c.

MikeMcC399:
- Workflows in [nodejs/corepack](https://github.com/nodejs/corepack) (currently using `v4`) will benefit from this PR. See https://github.com/actions/setup-node/issues/686#issuecomment-3338114303

v-priyagupta108:
Hello,

As part of this PR, we introduce detection support for `devEngines.packageManager`, enabling automatic caching for npm dependencies. This improvement will be included in the upcoming `v5` patch release. We believe this enhancement will help streamline workflows and improve consistency across your projects.

@scottohara 👋 , thank you for your thoughtful and detailed feedback. Your explanation of the differences between the `engines`, `devEngines`, and `packageManager` fields has been extremely helpful.

Based on community feedback, we started support for `devEngines.packageManager` and `packageManager` to address the needs of teams relying on these fields.

We recognize that `engines.npm` is the officially recommended way to specify the required npm version, and your insights underscore the importance of aligning with community best practices. In response to ongoing feedback and broader requirements, we will plan to add detection for `engines.npm` in a future update.

Your feedback and suggestions are always appreciated. Thank you!

mrgrain:
> * **npm:** Caching is now automatically enabled by default if `package.json` contains a `packageManager` field set to `npm` and no explicit `cache` input is provided.
> * **pnpm & yarn:** Caching is disabled by default to avoid compatibility issues. Users who wish to enable caching for `pnpm` or `yarn` can do so manually via the `cache` input.

This inconsistency between package managers just seems not worth it to me. 🤷🏻  Together with the concerns raised in https://github.com/actions/setup-node/issues/1358 I still suggest to just go back on automatic cache enablement.

kleinfreund:
@scottohara While I agree that `pkg.packageManager` is the wrong thing to use to infer whether npm is used (the field doesn't mean anything to npm), I disagree with your analysis on why `pkg.devEngines` shouldn't be used as part of the inference logic that's being discussed here.

> The key phrasing here seems to be _"...people interacting with the source code of a project"_.

> ❌ devEngines.packageManager would be inappropriate to use in this context, as setup-node does not exactly fit the definition of "...people interacting with the source code of a project"

That _is_ actually what GitHub Actions workflows almost categorically do when we're talking about projects based on Node.js. A workflow typically checks out a GitHub repository (that's what source code means here) to perform some tasks (e.g. run the project's linters or test suite(s)). To optimize for runner storage and workflow execution time, caching `node_modules/` and package manager things is useful. And that generally happens in the context of a project's source code. I'm not saying a workflow never `npm install`s a package that is being published through that repository (in which case you do look at `pkg.engines` and not `pkg.devEngines`, but that shouldn't be the common case. In fact, I'd argue that is so rare as to not consider it here.

So, `pkg.devEngines` is in my opinion definitely the more appropriate field to look at. More reliable still is whether `package-lock.json` exists next to `package.json`. `pkg.engines` is fine as a fallback, but for operations dealing with source code, `pkg.devEngines` should be considered the correct modern replacement.

> setup-node does seem to fit more with the definition of "...capable of properly installing your program".

In that documentation, installing your program refers to running `npm install your-program` or running `npm install` on a project that already contains your-program as a dependency. In other words, `pkg.engines` is about what Node.js version is supported when executing/running the program and as such is checked during installation of it as a dependency in another project.

alexaka1:
@scottohara 

That is not how I interpret the `engines` field. The details are lost in the fine print.

> Unless the user has set the engine-strict config flag, this field is advisory only and will only produce warnings when your package is installed as a dependency.

This field is primarily used for consuming packages. Ie. I distribute a package called `foobar` on NPM, and anyone who wants to install it must conform to MY `engines` property. I can give suggestions to downstream consumers on what version of node my package runs on. I have never seen this used for standardizing tool version in a dev team. This has always been used as a lower and upper bound for packages. But this hasn't been too relevant since TypeScript went viral.

`packageManager` is a Corepack feature. To me, this is the standardized tool versioning property. It can define the package manager version and automatically install if it is not installed.

`devEngines` was supposed to be the replacement for Corepack after Node decided to kill it. 

So if you're gonna look at any property, it should be `devEngines` as that is what is supposed to control tools versioning (such as in CI). `engines` should NOT be used at all.

Julusian:
^ Adding some examples to this, in most of my yarn based projects, the engines is either not set or only includes node version ranges.

The only one of the 3 properties that is set is `packageManager`, but as that is a corepack property, once/if yarn makes their own entrypoint for yarn, I would expect to see that property disappear. Maybe they will use devEngines instead, or maybe they will add something to their .yarnrc.yml file.
So to be futureproof, any deetection should not be trying to rule out other package managers and assuming npm, but instead confirm that the project is npm 

I also have an unusual case to consider, in one project I have:
```
"engines": {
    "npm": "please-use-yarn",
    "yarn": "^4.5",
    "node": ">=22.19.0 <23"
  },
```
because inexperienced users kept not reading the docs and trying to use npm then ask for help with the confusing output. It wouldnt be terrible if this was misclassified, but it would be nice to not have to fight the action on this

kachkaev:
> This improvement will be included in the upcoming v5 patch release. 
> _from https://github.com/actions/setup-node/pull/1374#issuecomment-3339010276_

Please release it as a new major (v6). See https://github.com/actions/setup-node/pull/1374#issuecomment-3319047932 for details.

scottohara:
> This inconsistency between package managers just seems not worth it to me. 🤷🏻 Together with the concerns raised in #1358 I still suggest to just go back on automatic cache enablement.

Given the general disagreement above on the purpose and intent of the various different fields and ways a package manager might be specified (by the author) or inferred (by this action), I tend to agree that this seems to be causing more issues than it solves.

Unless automatic cache enablement could be made to work flawlessly with any package manager, and without requiring specific fields in the manifest, then a return to manual cache enablement seems like the best path forward here (that is, return to v4 behaviour).

MikeMcC399:
@priyagupta108 / @scottohara 

> Unless automatic cache enablement could be made to work flawlessly with any package manager, and without requiring specific fields in the manifest, then a return to manual cache enablement seems like the best path forward here (that is, return to v4 behaviour).

Perhaps this needs to be a separate issue / proposal? It seems that this PR is going ahead with enabling caching for npm by default, despite the arguments against this.

README.md
```diff
@@ -12,10 +12,14 @@ This action provides the following functionality for GitHub Actions users:
 - Registering problem matchers for error output
 - Configuring authentication for GPR or npm
 
+## Breaking changes in V6
+
+- Caching is now automatically enabled for npm projects when the `packageManager` field in `package.json` is set to `npm`. For other package managers, such as Yarn and pnpm, caching is disabled by default and must be configured manually using the `cache` input.
```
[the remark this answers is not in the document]
MikeMcC399: This comment is now resolved. Thank you!

futursolo:
In my opinion, since both yarn and pnpm essentially recommend corepack as the recommended way to install them,
it would make sense to simply honour the packageManager field via corepack even after Node 25 removes it by default.

This is regardless of whether caching behaviour is automatic or manual.

The last couple comments in #1357 summarised this issue quite well.

Without built in corepack support (or package manager versioning support), workflows with yarn / pnpm basically becomes:

setup-node -> corepack / manually install package manager -> setup-node (cache)

This is a highly undesirable approach for anyone using yarn or pnpm.
Hence, setup-node should manage this internally if it tries to interact with yarn or pnpm.

As for how to check what package manager is used, the action can simply let corepack decide what is the package manager with `corepack install`.

If the caching is manual, then package managers is already specified and this action can simply call `corepack yarn/pnpm`.

.github/workflows/e2e-cache.yml
```diff
@@ -107,7 +107,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
```
MikeMcC399, line 109: Is it time to also drop `macos-13`, since this is now deprecated?

.github/workflows/versions.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
-        node-version: [20-nightly, 21-nightly, 18.0.0-nightly]
+        node-version: [20-nightly, 21-nightly, 24.0.0-nightly]
```
MikeMcC399, line 85: Node.js 21 reached EOL on Jun 1, 2025
Should this be changed to `22-nightly`?

.github/workflows/e2e-cache.yml
priyagupta108: Updated the workflows.

mrgrain:
> @priyagupta108 / @scottohara
> 
> > Unless automatic cache enablement could be made to work flawlessly with any package manager, and without requiring specific fields in the manifest, then a return to manual cache enablement seems like the best path forward here (that is, return to v4 behaviour).
> 
> Perhaps this needs to be a separate issue / proposal? It seems that this PR is going ahead with enabling caching for npm by default, despite the arguments against this.

IDK. This has already been raised in a number of issues and now on this PR. Seems to me the maintainers are aware of the suggestion, but have prioritised a different goal (automated caching for npm) as more important than any of the raised concerns. Which is absolutely their prerogative to do.

Examples:
https://github.com/actions/setup-node/issues/1357#issuecomment-3339836539
https://github.com/actions/setup-node/issues/1358#issuecomment-3259195877

v-priyagupta108:
@MikeMcC399, Thank you for your feedback.

For this PR, the current approach is based on the implementation of feature request #686 and the feedback received so far. However, we’ll continue to monitor for additional feedback that may arise.

If you feel strongly about this or have further suggestions, please consider opening a dedicated issue or proposal so the discussion can be more focused and visible to the community.

@futursolo ,
Thank you for sharing your insights! We’ve received similar feedback in other discussions, such as #531 and we understand the value this could add to supporting diverse project needs.
We’ll look into this and explore how to incorporate these improvements in future updates, while ensuring compatibility and maintainability.
In the meantime, if you have any additional suggestions or use cases, please feel free to share them. Your feedback is greatly appreciated!

MikeMcC399:
@priyagupta108 

> If you feel strongly about this or have further suggestions, please consider opening a dedicated issue or proposal so the discussion can be more focused and visible to the community.

Thanks for picking up the individual suggestions I made for this PR! I've reviewed the usage of `setup-node` in the workflows I'm interested in and this PR will work for me, so at this time I don't have a dedicated issue / proposal that I plan to submit.

I look forward to integrating the `v6` version when you release it!

priya-kinthali (approved):

steveoh:
Out of curiosity, what pnpm compatibility issues were identified with caching? Was this only related to the automatic caching or also with the cache: pnpm argument as well?

Issue #663, opened 2023-01-10
Add option to restore cache only


ForsakenHarmony:
**Description:**
It would be nice to have the option to restore only and not save.

**Justification:**
In bigger repos, it's easy to hit the 10gb cache limit. Not caching in PRs can help with this.
Right now, I have to manually delete caches to help keep the caches on `main` alive.

**Are you willing to submit a PR?**

dmitry-shibanov:
Hello @ForsakenHarmony. Thank you for your report. We'll investigate the issue.

desrosj:
Found this searching for an open issue because I wanted to request the same feature. Some context on the specific use case I have.

In workflows triggered with the `pull_request_target` event, the recommendation in the docs is as follows:
> Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch. To help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered.

When using the `setup-node` action, I'd like to turn off cache saving for runs triggered with this event.

hugo082:
It can also help when a GitHub matrix is used. On large matrices, we may end with concurrency on cache write, forcing the job to wait (more than 1m on my repo).

```
Post job cleanup.
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/... --files-from manifest.txt --use-compress-program zstdmt
Failed to save: Unable to reserve cache with key <KEY>, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/main, Key: <KEY>, Version: <HASH>
```

maschwenk:
@ForsakenHarmony agree. In a big monorepo with 500+ open PRs and ~100 commits per day, the cache becomes unusable because there's too many circumstances where the `main` cache gets evicted because of hundreds of non-main caches. I can't decide if:

1. I need to control which bits of my pipeline are restore-only
2. I should automate purging all non-main caches

But it feels like a little of both are necessary

v-aparnajyothi-y:
Hello Everyone, Thanks everyone for your patience and contributions. This issue reported here has been addressed in [PR #1419](https://github.com/actions/setup-node/pull/1419), which is now merged. The documentation helps cache state restoration handling.
We are proceeding to close this issue as the related PR is merged, Please feel free to reach us in case of any further concerns to reopen this issue :)

ForsakenHarmony:
[...]

PR #1419, opened 2025-10-21
Add example for restore-only cache in documentation

aparnajyothi-y:
**Description:**
This PR adds documentation and an example workflow demonstrating how to achieve restore-only caching with setup-node using the actions/cache.
The update is based on community feedback from [#663](https://github.com/actions/setup-node/issues/663) (“Add option to restore cache only”), which proposed adding a restore-only flag to the action.
Instead of introducing new functionality, this documentation clarifies how users can already achieve this behavior using actions/cache.
**Related issue:**
[#663](https://github.com/actions/setup-node/issues/663)

**Check list:**
- [ X] Mark if documentation changes are required.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -327,7 +327,7 @@ jobs:
       - run: npm install
 ```
 
-> **Note**: Using a restore-only cache avoids redundant writes to the cache, which can reduce workflow time and minimize cache storage usage, as referenced in cache scenarios for [Node – npm](https://github.com/actions/cache/blob/main/examples.md#node---npm).
+> For more details related to cache scenarios, please refer [Node – npm](https://github.com/actions/cache/blob/main/examples.md#node---npm).
 
 ## Multiple Operating Systems and Architectures
 


gowridurgad (approved):

Mrquinn4 (approved):

HarithaVattikuti (approved):

Kas-tle:
Not sure why this was merged as this example will pretty much always fail. The key `${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}` does not match how this action formats its cache keys. Currently, they are formatted as:

```
node-cache-${{ runner.os }}-RUNNER_ARCH-PACKAGE_MANAGER-${{ hashFiles('**/yarn.lock') }}
```

Note that runner architecture and package manager must be specified manually as the latter is not provided as action context, and the former, though provided as `${{ runner.arch }}` returns uppercase letters instead of lowercase (as this action formats architecture).

Furthermore, the provided path will also fail as caches (at least on Linux runners) are usually stored under `/home/runner/...`, not `/home`. It is also worth pointing out that other package managers (e.g. yarn) would require an alternate path (e.g. `/home/runner/.yarn/berry/cache`).

Finally, the base action from `actions/cache` is not restore only, so the provided example would actually result in the cache being duplicated... `actions/cache/restore` should be used instead. See https://github.com/actions/cache/tree/main/restore.

PR #1434, opened 2025-11-12
Add GitHub Actions workflow for npm package publishing

mikelandkristinfox-byte:
This workflow runs tests and publishes a Node.js package to GitHub Packages upon release creation.

**Description:**
Describe your changes.

**Related issue:**

**Check list:**

### DIFF
--- a/.github/workflows/npm-publish-github-packages.yml
+++ b/.github/workflows/npm-publish-github-packages.yml
@@ -1,0 +1,36 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish-github-packages.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
ScottBrenner:
```suggestion
      - uses: actions/checkout@v6
      - uses: actions/setup-node@v6.2.0
```

v-gowridurgad:
Hi @mikelandkristinfox-byte, thanks for the contribution. After reviewing this, we decided not to add publishing support because this workflow would try to publish the same package version on subsequent runs unless package.json is updated first, which would cause the publish step to fail. So, we’re going to close this PR. Thanks for understanding.

PR #1240, opened 2025-03-06
feat: support private mirrors

marco-ippolito:
Fixes: https://github.com/actions/setup-node/issues/1150
Fixes: https://github.com/actions/setup-node/issues/730

PR #1467, opened 2026-01-14
fix: remove hardcoded bearer

marco-ippolito:
Followup of https://github.com/actions/setup-node/pull/1240

**Description:**
This change makes the mirror usable.
Right now Immagine we setup the value of `mirror-token` to `foo`.
In this line the value is used as
https://github.com/actions/setup-node/blob/2951748f4c016b747952f8ca7e75fc64f2f62b53/src/distributions/base-distribution.ts#L106

`Bearer foo` but in other places such as:
https://github.com/actions/setup-node/blob/2951748f4c016b747952f8ca7e75fc64f2f62b53/src/distributions/base-distribution.ts#L149

https://github.com/actions/setup-node/blob/2951748f4c016b747952f8ca7e75fc64f2f62b53/src/distributions/base-distribution.ts#L210

its used as is.

But if we change the value to `Bearer foo`,
https://github.com/actions/setup-node/blob/2951748f4c016b747952f8ca7e75fc64f2f62b53/src/distributions/base-distribution.ts#L106
here it becomes `Bearer Bearer foo` which is incorrect, but works everywhere else.

So this PR makes it that if its a `Bearer`, the user needs to be explicit.

**Related issue:**

**Check list:**

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -54246,7 +54246,7 @@ class BaseDistribution {
         const dataUrl = `${initialUrl}/index.json`;
         const headers = {};
         if (this.nodeInfo.mirrorToken) {
-            headers['Authorization'] = `Bearer ${this.nodeInfo.mirrorToken}`;
+            headers['Authorization'] = this.nodeInfo.mirrorToken;
         }
         const response = await this.httpClient.getJson(dataUrl, headers);
         return response.result || [];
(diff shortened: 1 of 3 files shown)


marco-ippolito:
Hi @aparnajyothi-y, @priyagupta108 can you please take a look

marco-ippolito:
friendly ping

aparnajyothi-y (approved):

HarithaVattikuti (approved):

Issue #1027, opened 2024-05-01
If `cache: yarn` is specified, this action fails


esperecyan:
**Description:**
If I set `packageManager` property in `package.json` and set `cache: yarn` in this action, this action will fail.

Related: 
- https://github.com/actions/setup-node/issues/531#issuecomment-1816774699

**Action version:**
v4 (v4.0.2)

**Platform:**
- [x] Ubuntu
- [ ] macOS
- [ ] Windows

**Runner type:**
- [x] Hosted
- [ ] Self-hosted

**Tools version:**
- yarn 4.1.1

**Repro steps:**  
This is the workflow that reproduced this issue with the minimum configuration.
https://github.com/esperecyan/actions-setup-node-debug/actions/runs/8907795048/workflow
You can confirm that the issue does not occur in the following cases:
- If I set `npm` to `packageManager`.
- If I set `yarn` to `packageManager` and do not set `cache: yarn` for this action.

**Expected behavior:**
`actions/setup-node` succeeds.

**Actual behavior:**
`actions/setup-node` fails with the following error.
> Error: error This project's package.json defines "packageManager": "yarn@4.1.1". However the current global version of Yarn is 1.22.22.
> 
> Presence of the "packageManager" field indicates that the project is meant to be used with Corepack, a tool included by default with all official Node.js distributions starting from 16.9 and 14.19.
> Corepack must currently be enabled by running corepack enable in your terminal. For more information, check out https://yarnpkg.com/corepack.
https://github.com/esperecyan/actions-setup-node-debug/actions/runs/8907795048/job/24462266566

Running `corepack enable` before `actions/setup-node` will avoid this issue.
https://github.com/esperecyan/actions-setup-node-debug/actions/runs/8907795048/job/24462266822

HarithaVattikuti:
Hello @esperecyan 
Thank you for creating this issue. We will investigate it and get back to you as soon as we have some feedback.

1kevgriff:
Hitting this today too.

n0099:
#901
#546
#480

SunsetTechuila:
> Running corepack enable before actions/setup-node will avoid this issue.

works only for node 18+

for node 14-16 see https://github.com/actions/setup-node/issues/1075#issuecomment-2165643488

unickq:
Still there, 2025.
```
Environment details
  node: v18.20.6
  npm: 10.8.2
  yarn: 1.22.22
/usr/local/bin/yarn --version
1.22.22
[...]

PR #1479, opened 2026-01-18
fix(#1027): Improve yarn v4+ corepack support with better error handling

Satishchoudhary94:
### Problem
Yarn v4+ requires corepack, but the action tried to use yarn v1.22.22 from the system,
causing the error:

### Solution
- Automatically enable corepack before checking yarn version (for Node 16.9+)
- Detect corepack errors and provide clear, actionable messages
- Users can either enable corepack first or disable caching

### Changes
- Added `enableCorepackIfSupported()` helper function
- Updated yarn `getCacheFolderPath()` to enable corepack and detect version mismatch errors
- Improved error messages with links to GitHub issue

### Testing
- Builds successfully without errors
- Error messages guide users to solution

### DIFF
--- a/dist/cache-save/index.js
+++ b/dist/cache-save/index.js
@@ -44172,7 +44172,24 @@ exports.supportedPackageManagers = {
         name: 'yarn',
         lockFilePatterns: ['yarn.lock'],
         getCacheFolderPath: async (projectDir) => {
-            const yarnVersion = await (0, exports.getCommandOutputNotEmpty)(`yarn --version`, 'Could not retrieve version of yarn', projectDir);
+            // Try to enable corepack first if available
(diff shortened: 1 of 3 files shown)


v-gowridurgad:
Hi @Satishchoudhary94, thank you for your contribution! 

After discussion, we’ve decided not to pursue this PR at this time because starting with Node.js 25, Corepack will no longer be bundled with Node.js, which affects how this change would work for future versions. For reference, please visit the [nodejs-corepack](https://r2.nodejs.org/download/release/latest-v22.x/docs/api/corepack.html) documentation.

If you have any further concerns, please don’t hesitate to reach out. We appreciate your feedback and suggestions for future enhancements.

PR #1508, opened 2026-03-09
Add cache-write input for read-only cache mode

salmanmkc:
When using `cache` in workflows triggered by `pull_request`, there's no way to benefit from cached dependencies without also writing back to the cache at the end. This opens the door to cache poisoning, a malicious PR could push tainted packages into the cache that then get used by trusted runs on main.

This adds a `cache-write` input (defaults to `true`, fully backward compatible). When set to `false`, the post-step cache save is skipped so the action still restores deps from cache but never writes.

**Usage:**
```yaml
- uses: actions/setup-node@v6
  with:
    node-version: 20
    cache: npm
    cache-write: ${{ github.event_name != 'pull_request' }}
```

**What changed:**
- `action.yml` — new `cache-write` input
- `src/cache-save.ts` — early return when `cache-write` is `false`
- `dist/` — rebuilt

Pairs with similar PRs across setup-python, setup-go, setup-java, setup-dotnet for a consistent experience.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -25,6 +25,9 @@ inputs:
     default: true
   cache-dependency-path:
     description: 'Used to specify the path to a dependency file: package-lock.json, yarn.lock, etc. Supports wildcards or a list of file names for caching multiple dependencies.'
+  cache-write:
+    description: 'Whether to save the cache at the end of the workflow. Set to false for cache read-only mode, useful for preventing cache poisoning from untrusted PR builds.'
+    default: true
   mirror:
     description: 'Used to specify an alternative mirror to download Node.js binaries from'
   mirror-token:
--- a/dist/cache-save/index.js
+++ b/dist/cache-save/index.js
@@ -71532,6 +71532,11 @@ process.on('uncaughtException', e => {
 // Added early exit to resolve issue with slow post action step:
 async function run(earlyExit) {
     try {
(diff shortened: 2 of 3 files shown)


brunoborges:
After reviewing the current GitHub Actions cache security model, I recommend closing this PR as unnecessary for the stated cache-poisoning threat.

- `pull_request` caches are scoped to `refs/pull/.../merge` and cannot be restored by the base branch or other pull requests.
- Low-trust events resolving to the default branch receive read-only cache access from the backend.
- Current `@actions/cache` releases honor backend cache modes and handle denied writes non-fatally.

The suggested `cache-write: ${{ github.event_name != 'pull_request' }}` pattern disables useful PR-local caching without adding protection for the default branch, and it remains true for `pull_request_target`, where backend enforcement is what provides protection.

Explicit restore-only behavior may still be useful ergonomically, but GitHub already recommends composing `actions/cache/restore` for that case rather than adding a security-specific setup-action input.

References:
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-access-for-low-trust-workflow-triggers
- https://github.com/actions/toolkit/blob/main/packages/cache/RELEASES.md

Issue #1445, opened 2025-12-03
Trusted Publisher documentation


jasonkarns:
**Description:**
Document an example of how publishing to npm is handled if using npm's Trusted Publisher setup. In particular, explicitly answer the question whether `NODE_AUTH_TOKEN` needs to be set at all if using Trusted Publisher. (Presumably no?)

**Justification:**
Npm is pushing folks towards Trusted Publisher setup and away from token auth. This action should acknowledge that reality and document an example of what is necessary (or not necessary!) when using Trusted Publisher. Presently, there isn't so much as a mention of Trusted Publisher setup or configuration impact.

**Are you willing to submit a PR?**

v-priyagupta108:
Hello @jasonkarns,
Thank you for this feature request. We will investigate it and get back to you as soon as we have some feedback.

jasonkarns:
Some additional bits that I would recommend in the docs:

The npm documentation uses node 20 as an example, but that node does _not_ ship with a compatible version of the npm CLI. (cli v11.5.1+ is required for the feature to work)

So whatever examples are used, should either show explicitly upgrading npm (if using the stock version of node which is presently v20); or better yet, give examples showing node v24 which includes a compatible version of npm OOTB and doesn't require additional setup.

mikehardy:
It's worse than whether it is unclear that NODE_AUTH_TOKEN needs to be set or not, it is the case that if you set up OIDC Trusted Publish to npmjs.org and you *don't* actually clear the NODE_AUTH_TOKEN that actions/setup-node makes up for you when it is unset, your publish will fail. And it will do so with a 404 error which is unhelpful (but understandable, as if you got a 401 from npmjs.org it would be considered a possible leak of existence status about private packages) 

Intensely frustrating. 

- #1440

mikehardy:
In the end I had to revert to using a granular token for an NPM publish. Something about the interaction of actions/setup-node, lerna-current (v9.0.3) and npm current (11.6.2 IIRC, was `npm i npm@latest` in the runner) was simply not working. Current versions of everything worked as recently as November 27 but failed starting then and I couldn't revive it despite an intense attempt.

I'll revisit in 3 months when my 90-day-max-duration token expires I suppose.

Shurtu-gal:
Our workflow still didn't work though any help would be appreciated. https://github.com/asyncapi/cli/actions/runs/20265856000/workflow

jasonkarns:
[...]

PR #1549, opened 2026-05-14
Explicitly disable caches in sensitive publish workflows

slorber:
**Description:**

Sensitive workflows like the Trusted Publishing workflow should explicitly disable cache to reduce supply chain risk due to cache poisoning. See recent hack: https://tanstack.com/blog/npm-supply-chain-compromise-postmortem

It would be better if that default was applied automatically, and not depending on various conditions that are not so easy to understand, so explicitly disabling this in the documentation is safer for now.

Automatic caching was afaik turned on by default in v5 in certain conditions. I think it's worth considering reverting that change and making caching opt-in, to have a more secure action by default, even if it's less performant.

Note that many people upgrade actions without looking at the changelog, using Dependabot. Default value changes like this can introduce vulnerabilities, so whatever the default is, explicitly disabling the cache for sensitive workflows remains a good idea IMHO.

Note that on README.md, there's already a disclaimer giving similar advice. However it's not easy to find and your Trusted Publishing docs don't even follow that recommendation:

CleanShot 2026-05-14 at 11 10 25@2x

https://github.com/actions/setup-node#breaking-changes-in-v5

I believe all your documented examples should follow that recommendation, explicitly disable caching with `package-manager-cache: false`, and link to a canonical doc page explaining the security problem.

This PR is just to kickstart the discussion; there are probably other sensitive workflows you document that do not have `package-manager-cache: false`, and probably other official GitHub Actions whose docs show insecure examples.

**Related issue:**

See my comment here: https://github.com/actions/setup-node/issues/1445#issuecomment-4428351441

**Check list:**
- [X] Mark if documentation changes are required.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -401,6 +401,7 @@ steps:
   with:
     node-version: '24.x'
     registry-url: 'https://registry.npmjs.org'
+    package-manager-cache: false # Prevent cache poisoning issues
 - run: npm ci
 - run: npm publish
   env:
@@ -408,6 +409,7 @@ steps:
 - uses: actions/setup-node@v6
   with:
     registry-url: 'https://npm.pkg.github.com'
+    package-manager-cache: false # Prevent cache poisoning issues
 - run: npm publish
   env:
     NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


v-chiranjib-swain:
Closing this PR as superseded by #1567.

PR #1592, opened 2026-07-21
Fix comment in tsconfig.json

AlbertoPimpo:
**Description:**
Just a small fix to the comment in `tsconfig.json`. The comment gives a list of options of `module`, but it's using one that is not part of the list. This PR adds this one in the list.

**Related issue:**
No need to create an issue for something this small.

**Check list:**


v-gowridurgad:
Hi @AlbertoPimpo, thanks for the PR. We reviewed it, and since the project already uses "module": "NodeNext", its validity is already clear. The comment here is only informational about possible options, and this change does not affect behavior. For that reason, we’re closing the PR.

PR #1627, opened 2026-09-03
fix: avoid GitHub API rate limits for the Node manifest

BridgeAR:
Large matrices can exhaust the repository GITHUB_TOKEN rate limit because each cold manifest lookup resolves a repository tree and blob through REST.

This reads the fixed raw manifest path first and caches the nonempty response for later setup-node steps in the job. check-latest always bypasses that cache. The existing authenticated API path remains the fallback.

Refs: https://github.com/DataDog/dd-trace-js/actions/runs/33776938250/job/100720885063?pr=9406

### DIFF
--- a/__tests__/official-installer.test.ts
+++ b/__tests__/official-installer.test.ts
@@ -14,6 +14,9 @@ import osm from 'os';
 import path from 'path';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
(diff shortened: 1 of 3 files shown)


BridgeAR:
@v-HarithaVattikuti @v-gowridurgad @v-priyagupta108 @v-chiranjib-swain PTAL, this would help us a lot :)

